PR #1, opened 2018-03-13
Pr/fix typos

hihuz:
A few typos I noted while reading the page.

Some changes are a bit subjective, as they were more rephrasing than actual typos

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -13,22 +13,22 @@ so we have a lot of things on our plate!
 Due to the nature of the product, our roadmap is doomed to evolve
 often according to the technologies around it.
 
-Here are examples:
+Here are a few examples:
 
-* a browser will support a new web API, making it relevant for
+* a browser now supports a new web API, making it relevant for
   production purpose
-* a library we use will release a new major version, bringing
+* a library we use releases a new major version, bringing
   awesomeness to our world
-* a solution we connect with will release a new version, with API
+* a solution we connect with releases a new version, with API
   changes (breaking, or allowing us to interact with it in a more
   convenient way)
 * what was once considered — at least by us — as a good practice
-  has now found a better alternative
+  has now been replaced by a better alternative
 
-While all of these things happens, **we have a promise to keep with you:
+While all of these things happen, **we have a promise to keep with you:
 you won’t have to suffer from this fatigue by using Front-Commerce.**
 
-Thus we will have some work to make it available in the product,
+Thus we will have to work to make it available in the product,
 without breaking your code or by making the upgrade as smooth as possible.
 
 ### We are a bootstrapped company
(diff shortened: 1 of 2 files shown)


docs/roadmap.md
```diff
@@ -13,22 +13,22 @@ so we have a lot of things on our plate!
 Due to the nature of the product, our roadmap is doomed to evolve
 often according to the technologies around it.
 
-Here are examples:
+Here are a few examples:
```
hihuz: I added "a few" here as it felt a bit better, this is not really necessary though

docs/roadmap.md
```diff
@@ -13,22 +13,22 @@ so we have a lot of things on our plate!
...
 
-* a browser will support a new web API, making it relevant for
+* a browser now supports a new web API, making it relevant for
```
hihuz: in the bullet points here, the last sentence is in present tense whereas the previous ones are in future tense.

It felt better to keep everything in present tense here but another way would be to put the last bullet point in future tense

docs/roadmap.md
```diff
@@ -13,22 +13,22 @@ so we have a lot of things on our plate!
...
-* a library we use will release a new major version, bringing
+* a library we use releases a new major version, bringing
   awesomeness to our world
-* a solution we connect with will release a new version, with API
+* a solution we connect with releases a new version, with API
   changes (breaking, or allowing us to interact with it in a more
   convenient way)
 * what was once considered — at least by us — as a good practice
-  has now found a better alternative
+  has now been replaced by a better alternative
```
hihuz: "has found" here sounded a bit strange to me, but this is debatable

docs/roadmap.md
```diff
@@ -13,22 +13,22 @@ so we have a lot of things on our plate!
...
 
-While all of these things happens, **we have a promise to keep with you:
+While all of these things happen, **we have a promise to keep with you:
 you won’t have to suffer from this fatigue by using Front-Commerce.**
 
-Thus we will have some work to make it available in the product,
+Thus we will have to work to make it available in the product,
```
hihuz: either this or "some work to do" would work here

docs/roadmap.md
```diff
@@ -48,13 +48,13 @@ leverage them.
 ### We are part of an ecosystem
 
 Truly valuing our Customers and Partners means that our roadmap will integrate
-your feedbacks.
+your feedback.
```
hihuz: I don't think feedback can take plural form

docs/roadmap.md
```diff
@@ -70,16 +70,16 @@ or a new tool for the documentation in our roadmap…
...
 >
 > But the cool thing is that we have a lot of traction these days and our partners have
-> several Front-Commerce projects in the pipe, with decisions in the next few weeks.
+> several Front-Commerce projects in the pipe, with decisions coming in the next few weeks.
 >
 > For all these reasons the Roadmap will evolve a lot by mid-April and we will
 > keep you updated.
 
-If you want to have an imprecise vision of what comes next here is a
+If you want to have an overview of what comes next here is a
```
hihuz: open for discussion, I didn't really like the term "imprecise" but I'm not sure here

docs/roadmap.md
real34: Good catch, thanks

docs/roadmap.md
real34: Wow, I learnt something… thanks :1st_place_medal:

docs/roadmap.md
real34: The thing is that it is less or more a guess… that being said it is very likely to be correct, so let’s use overview :metal:

PR #3, opened 2018-03-14
Second part — and end — of the vision page

real34:
Feedbacks welcome, there might be typos and things too long… it is a first try.

### DIFF
--- a/docs/vision.md
+++ b/docs/vision.md
@@ -78,16 +78,15 @@ knew most and which allows to support a lot of eCommerce use cases.
 **Front-Commerce is in production as a frontend for Magento2 stores since
 February 2018. It supports most of the Magento Opensource edition.**
 It is composed of a React application, a GraphQL middleware using Magento2
-webAPI, Redis for caching and ElasticSearch for search — more details on
+Web APIs, Redis for caching and ElasticSearch for search — more details on
 [Front-Commerce’s architecture documentation page](architecture-overview.md).
 
-We wanted to be our first integrators, and to create a few stores with
-Front-Commerce (dogfooding) before distributing the product to the world.
+We wanted to be our first integrators (dogfooding), and to create a few stores with
+Front-Commerce before distributing the product to the world.
 
-Being in production means that we focused on features, themes creation and
-overrides, payments, deployments, caching, logging… and everything you would
-expect from such a solution. **That makes us the most mature « eCommerce PWA theme »
-in our field.**
+Being in production means that we focused on features, themes, payments,
+deployments, caching, logging… and everything you would expect from such a
+solution. **That makes us the most mature « eCommerce PWA theme » in our field.**
 
 We are now working on technical documentation and refactoring (v1.0), along with
 business and legal things to start selling license to customers and help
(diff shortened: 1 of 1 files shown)


JulienPradet:
I've made a commit that fixes a few typos.
I'll follow up with a more general feedback comment.

docs/vision.md
```diff
@@ -72,12 +72,65 @@ moved to GraphQL and (the yet young) Apollo client to embrace their rich and
...
-## What do we want it to be
+We wanted to be our first integrators (dogfooding), and to create a few stores with
+Front-Commerce before distributing the product to the world.
+
+Being in production means that we focused on features, themes, payments,
+deployments, caching, logging… and everything you would expect from such a
+solution. **That makes us the most mature « eCommerce PWA theme » in our field.**
+
+We are now working on technical documentation and refactoring (v1.0), along with
+business and legal things to start selling license to customers and help
```
JulienPradet: There's a lot happening here. Maybe we should try to split it in two parts ?
1. Distribution ready
2. Legal/License/Partners ready

docs/vision.md
```diff
@@ -72,12 +72,65 @@ moved to GraphQL and (the yet young) Apollo client to embrace their rich and
...
+them building their stores with Front-Commerce. It also means that we are
+building an ecosystem of partners. See [the license page](/license.html)
+for further information.
 
-Proin eget tortor risus. Praesent sapien massa, convallis a pellentesque nec, egestas non nisi. Cras ultricies ligula sed magna dictum porta.
+**The product is alive, and ready for you to use!** We are now entering in a
```
JulienPradet: I get what you mean here, but I'm not sure that `for you` is necessary here.
Let's stay simple with `The product is alive, and ready to use!` as a product is always for its users.

docs/vision.md
```diff
@@ -72,12 +72,65 @@ moved to GraphQL and (the yet young) Apollo client to embrace their rich and
...
+
+* efficient collaboration between people and teams
+* web performance
+* accessibility
+* SEO
+* consistent web UIs
+* new web APIs
+* different types of tests (unit, integration, visual regression, contracts…)
+* continuous deployment
+* microservices architecture and resilient websites
```
JulienPradet: We'll change this later, but as a side note, I'd split the best practices in two :
* the ones for the users
* the ones for the developers (even if these are meant to enhance the experience of the users)

docs/vision.md
```diff
@@ -72,12 +72,65 @@ moved to GraphQL and (the yet young) Apollo client to embrace their rich and
...
+Our field evolves at a fast pace and **it is difficult to remain aware of all
+the best practices**. By using Front-Commerce, you will benefit from all our
+experience and get those new fancy practices with our regular updates:
+**your stores will perform better over time**.
+
+Relieved from this burden, we hope that **you will focus more on your
+users** and deliver the shopping experience they deserve and love.
+
+As developers, you will also learn new concepts and have everything you
+need to experiment what you never took the time to learn before.
```
JulienPradet: By telling them that they never took the time, you put the blame on them.
Maybe we can make it a bit softer : `what you wanted to learn but didn't yet` ?

docs/vision.md
```diff
@@ -72,12 +72,65 @@ moved to GraphQL and (the yet young) Apollo client to embrace their rich and
...
+**If you are working on a team and want to increase the overall quality,
+Front-Commerce can help you introduce better practices**.
+
+We believe that together we can build a faster and more usable web,
+starting with online stores. Are you ready for this revolution?
```
JulienPradet: Ideally we could add a call to action here. To the getting started/ receipe page ? Or the roadmap since the other pages are not ready yet.

docs/vision.md
hihuz: also there is a typo : license**s** (if the phrasing stays the same)

docs/vision.md
```diff
@@ -72,12 +72,65 @@ moved to GraphQL and (the yet young) Apollo client to embrace their rich and
...
+* SEO
+* consistent web UIs
+* new web APIs
+* different types of tests (unit, integration, visual regression, contracts…)
+* continuous deployment
+* microservices architecture and resilient websites
+
+Our field evolves at a fast pace and **it is difficult to remain aware of all
+the best practices**. By using Front-Commerce, you will benefit from all our
+experience and get those new fancy practices with our regular updates:
```
hihuz: Not sure about the term 'fancy' here, it sounds a bit negative / reductive to me, but I don't know what I would put in its place.

real34: I’ll simply remove it for now!

real34:
Thank you for your reviews. The page itself is quite long and will definitely needs to be reworked or splitted.

I am merging the current version, and we will be seeking for user feedbacks to understand what is important for us… the rest could be refactored as blog posts!

PR #16, opened 2019-01-23
fix "closed source product" url

gkueny:
change url "http://developers.front-commerce.test/en/license.html" to "https://developers.front-commerce.com/en/license.html"


real34:
Thanks a lot!

PR #22, opened 2019-02-19
Document how to extend the base theme

real34:


source/docs/essentials/extend-the-theme.md
```diff
@@ -1,6 +1,129 @@
 ---
+id: extend-the-theme
 title: Extend the theme
 ---
 
-> Content coming soon. See
-> [https://gitlab.com/front-commerce/front-commerce/issues/13](https://gitlab.com/front-commerce/front-commerce/issues/13)
+## Understanding theme overrides
+
+Front-Commerce web application is a fully featured e-commerce universal React application
```
JulienPradet: I'd add a link on the `universal` keyword explaining what this is about.

source/docs/essentials/extend-the-theme.md
```diff
@@ -1,6 +1,129 @@
...
+aimed at providing a sane base for your own theme.
+Even though you could reuse pages and components individually in a totally different theme,
+most of the time you might find easier and faster to start with the base theme and
+iterate from there.
+
+**Theme override is the mechanism allowing to extend, adapt, rewrite almost everything contained in the base theme.** It is what allows you to:
+- upgrade Front-Commerce and benefit from the latest component improvements,
```
JulienPradet: I'd add `customize the base theme` as a first item. It's close to the `extension` part but since we don't really have any extensions yet...

source/docs/essentials/extend-the-theme.md
```diff
@@ -1,6 +1,129 @@
...
+- customize how an extension is displayed,
+- or create a slightly different theme for the Black Friday in confidence!
+
+People having an understanding of [layout overrides in Magento](https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/layouts/layout-override.html), [child themes in Wordpress](https://developer.wordpress.org/themes/advanced-topics/child-themes/) / [Prestashop](https://devdocs.prestashop.com/1.7/themes/reference/template-inheritance/parent-child-feature/), [templates override in Drupal](https://www.drupal.org/docs/8/theming/twig/working-with-twig-templates), [themes in CakePHP](https://book.cakephp.org/3.0/en/views/themes.html) or other similar implementations should easily understand this mechanism in Front-Commerce.
```
JulienPradet: Would it make sense to talk about `template overrides in Magento 1` ?

Because if I understand the Magento2's doc correctly, it's only for the `layouts.xml`.

JulienPradet: I don't really know how to turn the phrase around, but I'm not a big fan of the current phrase because it tells people how they should feel.

Maybe something along those lines: "Having an understanding of ... will help you understand Front-Commerce's theme overrides because they all share the same spirit. But if you have no previous experience with these kind of features, let's see how it works!"

source/docs/essentials/extend-the-theme.md
```diff
@@ -1,6 +1,129 @@
...
+
+Your own theme will be located in its own folder and will use by default components from
```
JulienPradet:
```suggestion
Your own theme will be located in its own folder and will use default components from
```

source/docs/essentials/extend-the-theme.md
```diff
@@ -1,6 +1,129 @@
...
+an identical file structure.
+Your component will be used instead of the base one anywhere across the application.
+
+Themes are part of a module’s web area. Hence you must follow theses steps:
+
+1. define the module in the base configuration file
+2. copy the file (`js`, `scss` or `gql`) you want to override in the `theme` folder of your module
+3. customize its content in your module directory
+
+## Define the module in the base configuration file
```
JulienPradet: I'd change this name by using something like:

```suggestion
## Configure your custom theme and the parent theme(s) used in your application
```

Because it would make more sense according to [L22](https://github.com/front-commerce/developers.front-commerce.com/pull/22/files#diff-a821e827b1dd1b2089cc6ac3dfc6067bR22).

If you accept this suggestion, the title in the above list should be changed too.

source/docs/essentials/extend-the-theme.md
```diff
@@ -1,6 +1,129 @@
...
+- customize how an extension is displayed,
+- or create a slightly different theme for the Black Friday in confidence!
+
+People having an understanding of [layout overrides in Magento](https://devdocs.magento.com/guides/v2.3/frontend-dev-guide/layouts/layout-override.html), [child themes in Wordpress](https://developer.wordpress.org/themes/advanced-topics/child-themes/) / [Prestashop](https://devdocs.prestashop.com/1.7/themes/reference/template-inheritance/parent-child-feature/), [templates override in Drupal](https://www.drupal.org/docs/8/theming/twig/working-with-twig-templates), [themes in CakePHP](https://book.cakephp.org/3.0/en/views/themes.html) or other similar implementations should easily understand this mechanism in Front-Commerce.
+
+Your own theme will be located in its own folder and will use by default components from
+parent theme(s) — at least from Front-Commerce base theme.
+You would then copy the files you want to override in your theme folder by maintaining
+an identical file structure.
+Your component will be used instead of the base one anywhere across the application.
```
JulienPradet:
```suggestion
Your component will then be used instead of the one in your parent theme(s).
```

source/docs/essentials/extend-the-theme.md
```diff
@@ -1,6 +1,129 @@
...
+
+Themes are part of a module’s web area. Hence you must follow theses steps:
```
JulienPradet:
```suggestion
This translates in those three steps:
```

There no mention of module web area bellow. I'd keep it for the explanation within the second point.

source/docs/essentials/extend-the-theme.md
```diff
@@ -1,6 +1,129 @@
...
+
+1. define the module in the base configuration file
+2. copy the file (`js`, `scss` or `gql`) you want to override in the `theme` folder of your module
+3. customize its content in your module directory
+
+## Define the module in the base configuration file
+
+First, we need to create a minimalist module with the following file structure:
```
JulienPradet:
```suggestion
First, we need to create a minimalist module that will contain your own theme by creating the following file structure:
```

source/docs/essentials/extend-the-theme.md
```diff
@@ -1,6 +1,129 @@
...
+
+```
+my-module
```
JulienPradet: Wondering if it would make sense to mention that a module for now is supposed to be at the root of your application.

source/docs/essentials/extend-the-theme.md
```diff
@@ -1,6 +1,129 @@
...
+  url: "http://www.front-commerce.test",
+-  modules: [],
++  modules: ["./my-module"],
+  serverModules: [
+    // ...
+  ]
+};
+```
+
+Restart the application.
```
JulienPradet: Not sure if the point was to add a new line here, but if it was, there are spaces missing.

source/docs/essentials/extend-the-theme.md
```diff
@@ -1,6 +1,129 @@
...
+That’s it! You have successfully registered the module in your Front-Commerce application.
+
+## Override a component
+
+Let's add the description of a `Product` to a `ProductItem` as an example of overriding the base theme.
+
+The original file is: `node_modules/front-commerce/src/web/theme/modules/ProductView/ProductItem/ProductItem.js`
```
JulienPradet: Should we add an "info" blockquote to redirect to the folder structure explanation if one want to find their one component?

source/docs/essentials/extend-the-theme.md
```diff
@@ -1,6 +1,129 @@
...
+
+1. copy it to: `my-module/web/theme/modules/ProductView/ProductItem/ProductItem.js`
+2. add the description somewhere in your `ProductItem` with `{this.props.description}`.
+
+**However you will not see the description yet!**
```
JulienPradet: I understand that we want to avoid unnecessary restarts when overriding `ProductItem` but people might try to edit the component by adding a static string and it won't update.

```suggestion
**But you are not done yet!**
```

source/docs/essentials/extend-the-theme.md
```diff
@@ -1,6 +1,129 @@
...
+This information is not included in the GraphQL fields fetched by the application in the base theme.
+You will thus need to update the fragment related to `ProductItem`.
+
+The original fragment file is collocated with the original component at: `node_modules/front-commerce/src/web/theme/modules/ProductView/ProductItem/ProductItemFragment.gql`
+
+1. copy it to: `my-module/theme/molecules/Product/View/ProductItem/ProductItemFragment.gql`
```
JulienPradet:
```suggestion
1. copy it to: `my-module/web/theme/modules/ProductView/ProductItem/ProductItemFragment.gql`
```

source/docs/essentials/extend-the-theme.md
real34: Thanks!
I’ll go with « Configure your custom theme and use it in your application » for now.

PR #24, opened 2019-02-20
Document how to extend the GraphQL schema

real34:


source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,281 @@
...
+};
+```
+
+<blockquote class="info">
+**In depth:** under the hood, all module’s `typeDefs` are merged in a single array passed to graphql-tools' [`makeExecutableSchema`](https://www.apollographql.com/docs/graphql-tools/generate-schema.html#makeExecutableSchema) function
+</blockquote>
+
+### `resolvers` (optional)
+
+This is an object containing all the resolvers for resolving fields defined in the schema (usually from your `typeDefs`). This is where the implementation resides.
```
JulienPradet: Is it worth mentioning the types that are used as keys here (using Query as an example)? Even if it's documented in Apollo Tools I'd say it would make it easier for the readers not to switch context.

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,281 @@
...
+The `contextEnhancer` module definition key should be a function that initializes code that will be made available in GraphQL’s context, under the `loaders` key. **It is specific to Front-Commerce.**
+
+This is where you could construct models or [dataloaders](#TODO) that should be used in your GraphQL resolvers.
+
+The `contextEnhancer` function must return an object whose keys will be merged with previous modules’. It will be passed a single argument with the following keys:
+
+* `req`: the current server request
+* `loaders`: current loaders (from module initialized beforehand)
+* `makeDataLoader`: a factory to build a dataloader (see [dataloaders](#TODO))
+* `config`: the global configuration
```
JulienPradet: It's a bit weird to talk about `globalConfiguration` because I would expect to have access to files in `src/config` for instance. Maybe we could at least talk about `the server configuration`?

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,281 @@
...
+### `contextEnhancer` (optional)
+
+The `contextEnhancer` module definition key should be a function that initializes code that will be made available in GraphQL’s context, under the `loaders` key. **It is specific to Front-Commerce.**
+
+This is where you could construct models or [dataloaders](#TODO) that should be used in your GraphQL resolvers.
+
+The `contextEnhancer` function must return an object whose keys will be merged with previous modules’. It will be passed a single argument with the following keys:
+
+* `req`: the current server request
+* `loaders`: current loaders (from module initialized beforehand)
```
JulienPradet: It might be worth mentioning that it is important to register the correct dependency if one is using a loader from a previous module.

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,281 @@
...
+
+### `contextEnhancer` (optional)
+
+The `contextEnhancer` module definition key should be a function that initializes code that will be made available in GraphQL’s context, under the `loaders` key. **It is specific to Front-Commerce.**
+
+This is where you could construct models or [dataloaders](#TODO) that should be used in your GraphQL resolvers.
+
+The `contextEnhancer` function must return an object whose keys will be merged with previous modules’. It will be passed a single argument with the following keys:
+
+* `req`: the current server request
```
JulienPradet: Should we add a link to express docs here?

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,281 @@
...
+* `loaders`: current loaders (from module initialized beforehand)
+* `makeDataLoader`: a factory to build a dataloader (see [dataloaders](#TODO))
+* `config`: the global configuration
+
+Here is an example:
+
+```
+const MessageLoader = require("./loader");
```
JulienPradet: It might be good to add an actual implementation for MessageLoader? Especially since there's the weird (but awesome) constructor with currying.

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,281 @@
...
+  resolvers: {
+    Query: {
+      sayHello: (_, { name }, { loaders }) => {
+        return loaders.Message.load(`Hello ${name}`);
+      }
+    }
+  }
+  contextEnhancer: ({ req, loaders, makeDataLoader, config }) => {
+    return {
+      Message: MessageLoader(makeDataLoader)(config.apiBaseUrl)
```
JulienPradet: I think that this configuration does not exist in the core.

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,281 @@
...
+}
+```
+
+## Grouping small modules in a « meta-module »
+
+We recommend to keep your GraphQL modules small and focused on a single feature.
+If several teams are working on the project, small modules are a way to [federate the graph implementation](https://principledgraphql.com/integrity#2-federated-implementation).
+It also makes it easier to change implementations and service providers later in the project.
+For instance you could replace a `Magento2/Cms` module with a `Wordpress/Cms` one without impacting
+the web application (as soon as both schema shares a common API).
```
JulienPradet: Not sure about this one but:

```suggestion
the web application (as long as both schema shares a common API).
```

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,281 @@
...
+-    { name: "FrontCommerceCore", path: "server/modules/front-commerce-core" }
++    { name: "FrontCommerceCore", path: "server/modules/front-commerce-core" },
++    { name: "Magento2", path: "server/modules/magento2" }
+  ]
+};
+```
+
+Under the hood, it imports several modules at once (`Magento2/Cms`, `Magento2/Catalog`, `Magento2/Checkout`, `Magento2/Search`…).
+
+For a module to be considered as a meta-module, you only need to expose a list of
```
JulienPradet:
```suggestion
For a module to be considered as a meta-module, you need to expose a list of
```

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,281 @@
...
+sub-modules in the `modules` key of the module definition (along with the namespace).
```
JulienPradet: We almost had an API like documentation previously but everything wasn't documented (no paragraph about the `modules` keys). I'm wondering if it can be confusing.

JulienPradet:
Final thought: I've noticed that you are using the `src` folder for the custom module, but I've noticed that it's `my-module` in the other documentation pages.

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
+Let’s now add some executable logic to our module.
+
+### Implement resolvers
+
+The GraphQL middleware relies on resolver functions to determine the data to
+return for a given field. This is where most of the « real work » is done, for
+instance by fetching remote datasources or transforming data.
+
+Resolvers are exposed using the `resolvers` key of the module definition. It
+should be a **Resolver map**: an object where each key is a GraphQL type name,
```
JulienPradet: I'd put a link on to `https://www.apollographql.com/docs/graphql-tools/resolvers` here too.

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
+
+<blockquote class="info">
+  **In depth:** under the hood, Front-Commerce is generating a
+  `.front-commerce/modules.js` file from this configuration which is loaded very
+  early in the server bootstrapping process. The name is used to import the
+  correct module and export it in an homogeneous list, used by the
+  `withGraphQLApi` express middleware.
+</blockquote>
+
+**Congratulations!** You now have a new custom module ready to enhance the data
```
JulienPradet: Shouldn't we say that it needs a server restart here?

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
++  typeDefs: typeDefs
+};
+```
+
+<blockquote class="info">
+  Webpack is in fact taking care of converting SDL into a Javascript object
+  using an appropriate loader.
+</blockquote>
+
+**Congratulations again!** You should now be able to see these new fields and
```
JulienPradet: Maybe I'd add a mention that no restart is needed here because the server is hot reloaded and it's **awesome**.

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
+The resolver function will use the `sku` key from its parent data and will
+return the the current counter value from its local state (defaulting to `0` if
+no clicks occured).
+
+It is important to understand that the `sku` is not a parameter provided by
+frontend developers in the Query. It comes from data returned by earlier
+resolvers that fetched a `Product`, no matter where in the graph (category,
+upsells, cart…).
+
+<blockquote class="info">
```
JulienPradet: Sometimes we use note, sometimes we use info :man_shrugging:

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
+Please note that it could have been implemented as a top level Query as well:
```
JulienPradet: In this note, it's a bit weird if you don't really have a GraphQL background. I'd add something saying that in this case, the sku is a parameter and will be available in the second parameter of your resolver.

Or, i'd leave this as is and move this note in the schema definition since it's not related to the resolvers right now.

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
+      counters.set(sku, currentValueOf(sku) + incrementValue);
+      return {
+        success: true
+      };
+    }
+  }
+};
+```
+
+Let’s analyze this code part by part to understand what is possible to achieve
```
JulienPradet:
```suggestion
Let’s analyze this code to understand what is possible to achieve
```

And I'd make the parts explicit by adding titles. This makes it easier to see the transitions between each parts.

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
+in a module.
```
JulienPradet: I'm not sure about the module part. We only explain the resolvers here, not what a module is capable of. I'd simply remove it.

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
+APIs, but we recommend to try limiting as much as possible the number of
+top-level queries in your projects to learn _thinking in GraphQL_.
+
+</blockquote>
+
+The second part of this resolver map is a `incrementProductCounter` mutation.
+Mutations in GraphQL are a way to modify server-side data (whereas queries allow
+to read data). GraphQL modules must declare mutations by extending the top-level
+GraphQL `Mutation` type. You can read more about Mutations in the
+[GraphQL’s Mutations documentation](https://graphql.org/learn/queries/#mutations).
```
JulienPradet: Would it be good to mention GET vs POST here? It was made the most sense during our trainings.

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
+      return {
+        success: true
+      };
+    }
+  }
+};
+```
+
+In this resolver, the first parameter is unused (because in a top level resolver
+no earlier resolvers have been run so there are no _parent data_). The second
```
JulienPradet:
```suggestion
In this resolver, the first parameter is unused because in a top level resolver there is no _parent data_.
```

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
+
+```js
+// …
+module.exports = {
+  Product: {
+    clicksCounter: ({ sku }) => currentValueOf(sku)
+  }
+  // …
+};
+```
```
JulienPradet: I liked the fact that you've put a reminder of the mutation SDL below. It might be a good idea to put the SDL about the clicksCounter field here too.

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
+        success: true
+      };
+    }
+  }
+};
+```
+
+In this resolver, the first parameter is unused (because in a top level resolver
+no earlier resolvers have been run so there are no _parent data_). The second
+parameter contains arguments passed to the mutation. These parameters are the
```
JulienPradet:
```suggestion
parameter contains arguments passed to the mutation. These arguments are the
```

If you use the `parameters` term here, there might be a confusion with the parameters of a the resolver.

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
+`(sku: String!, incrementValue: Int)` part of the schema definition declared
+earlier:
+
+```graphql
+extend type Mutation {
+  incrementProductCounter(sku: String!, incrementValue: Int): MutationSuccess
+}
+```
+
+The resolver uses the mandatory `sku` argument to fetch the current counter
```
JulienPradet: This paragraph feels a bit confusing to me. I don't really know how to express it in a better way.

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
+value from its local state and defines a new value incremented of the
+`incrementValue` optional argument. If the `incrementValue` has not been
+provided in the GraphQL request, it defaults to `1`.
+
+Finally, the resolver returns a value matching the `MutationSuccess`
```
JulienPradet: I'd add a TODO html comment to add a reference about it.

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
+  "data": {
+    "product": {
+      "sku": "WH09",
+      "clicksCounter": 0
+    }
+  }
+}
+```
+
+Let’s increment it by sending a Mutation to our server:
```
JulienPradet:
```suggestion
You can even increment the counter by sending a Mutation to your server:
```

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
+```graphql
+# http://localhost:4000/playground
+mutation {
+  incrementProductCounter(sku: "WH09", incrementValue: 2) {
+    success
+  }
+}
+```
+
+If it was successful, the above query must now return `2` as the `clicksCounter`
```
JulienPradet:
```suggestion
If it was successful, the above query must now return `2` int the `clicksCounter` field.
```

mutiline :)

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
+value.
+
+### Asynchronous resolvers
+
+In the previous example resolvers were synchronous, meaning that they returned a
+value that was immediately available. Most of the time, your resolvers will
+fetch data from a remote service and return them asynchronously using a
+`Promise`.
+
+Returning `Promise` from resolvers is supported out of the box. We encourage you
```
JulienPradet: I don't see the value of the `We encourage you to experiment with it.` sentence.

There's quite no use of experimenting with hand made Promise here. I'd be better to talk about experimenting with an API imo.

source/docs/reference/graphql-module-definition.md
```diff
@@ -0,0 +1,181 @@
...
+The simplest (but useless) GraphQL module definition only requires an unique `namespace` key:
+
+```js
+// A minimal Front-Commerce GraphQL module that basically does nothing
+module.exports = {
+  namespace: "Acme/HelloWorld"
+};
+```
+
+This `namespace` can be used by other modules as a dependency, to ensure another module has been registered in the application.
```
JulienPradet: Might be a good thing to add an anchor to the `dependencies` field here.

source/docs/reference/graphql-module-definition.md
```diff
@@ -0,0 +1,181 @@
...
+
+<blockquote class="info">
+**Dependency management:** submodules dependencies will be resolved independently.
+
+It means that if a meta module `A` declares `A1` and `A2` in its `modules` key,
+and `A1` depends on module `B` then there is no guarantee that `A2` will be
+initialized _after_ `B`.
+
+You should make dependencies explicit either in the meta module definition or in
+each submodule definition.
```
JulienPradet: I have noticed there is no new lines created in the actual blockquote.

I'd add also an explanation about our reasoning here. Something like: Our goal is to make sure that each module can be used separatly. Thus, each dependency should be added to the module needing it and not in a meta module.

source/docs/reference/graphql-module-definition.md
```diff
@@ -0,0 +1,181 @@
...
+```js
+module.exports = {
+  namespace: "Acme/HelloWorld",
+  dependencies: ["Acme/Core"],
+  // …
+};
+```
+
+<blockquote class="warning">
+**Note:** we recommend to try to avoid dependencies between your modules. Usually this is a smell that the schema may be implemented differently and modules are too coupled together. A good use case is for instance to depend on a Front-Commerce core module.
```
JulienPradet: Is this really a smell? Having many is a smell, but there's almost no chance that a module has no dependency if we're thinking in Graph. For instance, you'll need to add your new field to `Product`, `Category` or anything else. It would not make sense most of the time to add it to the root query.

source/docs/reference/graphql-module-definition.md
```diff
@@ -0,0 +1,181 @@
...
+
+<blockquote class="info">
+**In depth:** modules are sorted using the [toposort](https://www.npmjs.com/package/toposort) library. See [`flattenAndReorderModulesUsingDependencies`](https://gitlab.com/front-commerce/front-commerce/blob/develop/src/server/core/graphql/__tests__/flattenAndReorderModulesUsingDependencies.js) code and tests for further details.
+</blockquote>
+
+## `typeDefs` (optional)
+
+GraphQL type definitions provided by this module. As a developer, this is a contract with the rest of your codebase.
+
+Type definitions must be [GraphQL schema language](https://graphql.org/learn/schema/) strings, and can be declared inline or loaded from a `.gql` file.
```
JulienPradet: Not sure about this one.

```suggestion
Type definitions must be [GraphQL Schema Definition Language](https://graphql.org/learn/schema/) strings, and can be declared inline or loaded from a `.gql` file.
```

source/docs/reference/graphql-module-definition.md
```diff
@@ -0,0 +1,181 @@
...
+
+<blockquote class="info">
+This is where you could construct models or [dataloaders](#TODO) that should be used in your GraphQL resolvers. See [Slim down resolvers with loaders](/docs/advanced/graphql/slim-down-resolvers-with-loaders.html) for more information.
+</blockquote>
+
+The `contextEnhancer` function must return an object whose keys will be merged with previous modules’. It will be passed a single argument with the following keys:
+
+* `req`: the current server request
+* `loaders`: current loaders (from module initialized beforehand)
+* `makeDataLoader`: a factory to build a dataloader (see [dataloaders](#TODO))
```
JulienPradet: I don't know if we'll merge before dataloaders, but if so this should be commented.

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,281 @@
...
+  serverModules: [
+    { name: "FrontCommerceCore", path: "server/modules/front-commerce-core" },
+-    { name: "Magento2", path: "server/modules/magento2" }
++    { name: "Magento2", path: "server/modules/magento2" },
++    { name: "HelloWorld", path: "./src/server/modules/hello-world" }
+  ]
+};
+```
+
+The `name` key must be unique across your server modules. It is a temporary name that is used during a code generation step and has no other usage in the application.
```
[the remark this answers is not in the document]
real34: It is only used in the generated `.front-commerce/modules.js` to name the import.

```js
import FrontCommerceCore from "server/modules/front-commerce-core";
import Magento2 from "server/modules/magento2";

module.exports = [FrontCommerceCore, Magento2];
```

I will create an issue to remove it by 1.0 and prevent the need for a name in the declaration (and prevent confusion).

source/docs/essentials/extend-the-graphql-schema.md
```diff
@@ -0,0 +1,443 @@
...
+};
+```
+
+<blockquote class="info">
+  Webpack is in fact taking care of converting SDL into a Javascript object
+  using an appropriate loader.
+</blockquote>
+
+**Congratulations again!** You should now be able to see these new fields and
+sue them in GraphQL.
```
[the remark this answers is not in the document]
real34: Yes, it should be `use` ;)

```suggestion
use them in GraphQL.
```

source/docs/essentials/extend-the-graphql-schema.md
real34: Well spotted… I tried to be careful and caught several mistakes but not this one.

source/docs/essentials/extend-the-graphql-schema.md
real34: I’ll try with bullet points.

source/docs/reference/graphql-module-definition.md
real34: There is a lot to say, and I agree that there are valid use cases. Let’s remove it for now and we will see how people use dependencies ;)

source/docs/advanced/graphql/slim-down-resolvers-with-loaders.md
```diff
@@ -0,0 +1,351 @@
...
+several advantages:
+
+- you can reuse existing code or libraries from another project
+- if you migrate away from Front-Commerce, this code could be reused easily in
+  your new application (as soon as it is in a Javascript environment of course)
+- testability: the code from both your resolvers and loaders would be easier to
+  test. You could inject fakes or mock loaders to test resolvers, and you could
+  test loaders without the complexity of a GraphQL context since they are pure
+  JS modules.
+- code sharing: under some circumstances you may also find convenient to share
```
JulienPradet: Is that for loaders or other things like adapters?  
Because here I understand it's about loaders, but it would mean we need to create a dataloader on the client side, etc.

I'm not quite sure this is the best idea.

source/docs/advanced/graphql/slim-down-resolvers-with-loaders.md
```diff
@@ -0,0 +1,351 @@
...
+          success: true
+        }));
+    }
+  }
+};
+```
+
+### Using GraphQL context for Dependency Injection
+
+In its current state, resolvers are tightly coupled to the current
```
JulienPradet: It's not really clear what current state you are refering to. I'd prefer something like

```suggestion
However, in the above example, resolvers are tightly coupled to the current
```

source/docs/advanced/graphql/slim-down-resolvers-with-loaders.md
```diff
@@ -0,0 +1,351 @@
...
+implementation. We recommend to leverage GraphQL’s context to inject a loader to
+resolvers.
```
JulienPradet:
```suggestion
resolvers in order to resolve this issue.
```

source/docs/advanced/graphql/slim-down-resolvers-with-loaders.md
```diff
@@ -0,0 +1,351 @@
...
++      return loaders.Counter.incrementBySku(sku, incrementValue)
+        .then(() => ({
+          success: true
+        }));
+    }
+  }
+};
+```
+
+**You might be wondering why we used the `loaders` value from the context, and
```
JulienPradet: I'd put the contextEnhancer before this refactoring. Because this paragraph looks like a... doc smell :thinking: 

I mean that by this that it's not explicit why would people be wondering this. Even though it seems clear to us.

source/docs/advanced/graphql/slim-down-resolvers-with-loaders.md
```diff
@@ -0,0 +1,351 @@
...
++    };
++  }
+};
+```
+
+<blockquote class="info">
+  Please refer to the [GraphQL module definition
+  reference](/docs/reference/graphql-module-definition.html#contextEnhancer-optional)
+  for the exact signature and a more complex example.
+</blockquote>
```
JulienPradet: I had an "A-HA!" moment when I understood that loaders could use existing loaders.
It might be a good idea to mention this in a blockquote and to talk about dependencies.

edit: reading further, it might make sense to add it in the WIP section

source/docs/advanced/graphql/slim-down-resolvers-with-loaders.md
```diff
@@ -0,0 +1,351 @@
...
+import resolvers from "./resolvers";
++import CounterLoader from "./loader";
+
+export default {
+  namespace: "ClicksCounters",
+  typeDefs: typeDefs
+  typeDefs: typeDefs,
+-  resolvers: resolvers
++  resolvers: resolvers,
++  contextEnhancer: () => {
```
JulienPradet: I'd set an unused option/parameters argument to show explicitly that there are some here.

source/docs/advanced/graphql/slim-down-resolvers-with-loaders.md
```diff
@@ -0,0 +1,351 @@
...
+you to get used to this pattern and explore the following documentation pages to
+see how you could combine them to achieve your goals.
+
+<blockquote class="wip">
+  The content below is currently being written. If you need more detailed
+  information, please [contact us](mailto:contact@front-commerce.com). We will
+  make sure to answer you in a timely manner.
+</blockquote>
+
+- optimize remote data fetching with dataloaders
```
JulienPradet: I think that this list could be within the blockquote.

source/docs/essentials/extend-the-graphql-schema.md
real34: We’ll have to homogenize it when/if styling it. For now, I agree: :woman_shrugging:

source/docs/essentials/extend-the-graphql-schema.md
real34: I’ll move it in the GraphQL part. It is more relevant there indeed.

source/docs/advanced/graphql/slim-down-resolvers-with-loaders.md
real34: It was more about the fact that loaders were pure JS code: it could lead to refactoring parts in smaller modules that could be shared.

This is not the best point of all of them. We will see if it leads to problematic implementations or poor practices.

source/docs/advanced/graphql/slim-down-resolvers-with-loaders.md
real34: I disagree on this one, even though I am not convinced either by the transition.

IMO it is really important to point out how the direct import in resolvers (that were just created above) creates coupling. Also I feel that developers will have a basic knowledge of DI and might see a point in this refactoring, while knowing that there is a « kind of DI container somewhere » doing some dark magic from an obscure YAML or XML configuration file.

I’ll keep it for now, but let’s change it if we see some incomprehension from readers…

source/docs/advanced/graphql/slim-down-resolvers-with-loaders.md
real34: @JulienPradet sorry, I did not get it… could you please do it or explain your thoughts further?

PR #37, opened 2019-02-27
Add WIP multiple store explanation

JulienPradet:


theme/layout/page.ejs
```diff
@@ -25,7 +25,15 @@
               <div class="subtitle-page">
                 <%- page.description %>
               </div>
-              <% } %>
+            <% } %>
+            <% if (page.wip) { %>
+              <blockquote class="wip">
+                <strong>Work in progress:</strong> This documentation is still a work in progress.
```
real34:
```suggestion
                <strong>Work In Progress:</strong> This documentation is still a work in progress.
```

theme/layout/page.ejs
```diff
@@ -25,7 +25,15 @@
...
+                We apologies for the inconvenience. If you want to learn more this subject, please
+                feel free to <a href="mailto:contact@front-commerce.com">contact us</a> or to upvote
+                the issue within the <a href="<%- page.wip %>">related gitlab issue</a>.
```
real34: I think we should migrate issues on Github after we’ve released the v2. Thus we would keep issues close to the code, plus it does not make any sense to include documentation issues in FC’s milestones IMO.

```suggestion
                the issue within the <a href="<%- page.wip %>">related issue</a>.
```

JulienPradet: Agreed, but some may not be documentation issues but feature issues.

PR #38, opened 2019-02-27
Add references about configurations and environment variables

JulienPradet:
Two things to note here :

* As discussed together, I've renamed the `REACT_APP_` env since nobody is using it yet. It still needs to be done in the core though.
* I've put the environment variables about Payments in order to document them. As discussed together, this will be moved once we've added the relevant documentations about the payments method.

### DIFF
--- a/source/docs/reference/configurations.md
+++ b/source/docs/reference/configurations.md
@@ -1,0 +1,107 @@
+---
+id: configurations
+title: Configurations
+---
+
+<blockquote class="note">
+This documentation is the current state of the configurations available in Front-Commerce. However, our goal is to clean it up to make sure that it is still easily customizable, but also easier to check if your configurations are correct. We will provide a clear migration path and will integrate this change within our 1.0.0 release.
+
+Please refer to the [related gitlab issue](https://gitlab.com/front-commerce/front-commerce/issues/106) to follow its advancement.
+</blockquote>
+
+## How to set your configurations
+
+Each file described below exist in [`node_modules/front-commerce/src/config`](https://gitlab.com/front-commerce/front-commerce/tree/develop/src/config) in Front-Commerce. If you want to override some, dupplicate those in `my-module/config`.
+
+## `config/website.js`
+
+This configuration file should contain any thing that impacts the content of your website. The term website refers to what a [`website` is in Magento's ecosystem](https://devdocs.magento.com/guides/v2.3/config-guide/multi-site/ms_over.html).
+
+* `root_categories_path` (ex: `1/517/`): which category to use for the main navigation menu. It will then be the children of this category that will be displayed.
+* `default_image_url` (ex: an absolute URL of an image): which image to use when no image path has been given to `<ResizedImage>`
+* `defaultTitle`: the default meta title of your application
+* `defaultDescription`: the default meta description of your application
+* `available_page_sizes`: which page sizes to display in a product list page (or any page with a pagination)
+* `website_id`: which website is used within magento (needed for customer related mutations)
+* `tax` (ex: `1.2` for 20% VAT): Used to correctly filter products within layer queries (this configuration should soon be deprecated with search updates in Front-Commerce)
+* `contentSecurityPolicy`: For security reasons only URLs from the store's domain are authorized through CSPs. However, for tracking and external dependencies, we may authorize more domains. Use the following config and add your custom domain in each :
+```
+{
+  directives: {
+    scriptSrc: [],
+    frameSrc: [],
+    styleSrc: [],
+    imgSrc: [],
+    fontSrc: [],
+    connectSrc: [],
+    baseUri: []
+  }
+}
+```
+  When a CSP is missing, your browser will let you know in its console.
+* `search`: an object that defines how the Elasticsearch queries should be made.
+  * `dynamicFacetSize`: number of filters to fetch from Elasticsearch
+  * `fixedFacets` (available values: `["max_price", "min_price"]`): filters that are not returned by default by Elasticsearch
+  * `blacklistKeys`: filters that are returned by Elasticsearch but we don't need to display
+  * `categoriesField` (default: `categories`): the field name that contains the category related to a product
+  * `authorizedCategoriesFacet`: array of ids of the categories that should be displayed in the filters
+* `phoneNumber`: support/contact number of the website
+* `email`: support/contact email of the website
+* `maxAddressLength`: max length for one line of address ([default is 35](https://webarchive.nationalarchives.gov.uk/+/http://www.cabinetoffice.gov.uk/media/254290/GDS%20Catalogue%20Vol%202.pdf))
+* `rewrittenToRoot`: array of the URLs that should be redirected to `/`
+
+You could find many other configurations in such a file because some configurations could come from optional modules. Some of these are:
+* `color_attribute_id`: allows to display the colors swatches for the color attribute (swatches should be detected automatically in the future)
+* `mapsKey`: Google maps key for Colissimo shipping method
+
+### `config/licences.js`
+
+This makes it possible to configure the GSC text.
+
+This file should export an object where the keys are the store's code (see [Configure multiple stores](/docs/advanced/production-ready/multistore.html)) and the value is a string (with some html content if needed).
+
+### `config/httpAuth.js`
+
+Configure a [Basic Authentication](https://en.wikipedia.org/wiki/Basic_access_authentication) on your server (usually for dev environments).
+
+```js
+module.exports = {
+  // enable basic authentication
+  enable: true,
+  // allow these IPs to bypass the basic authentication
+  whitelist: [
+    "x.x.x.x",
+  ]
+
+  // Headers where it's most likely to find the client IP
+  // Useful for logging the client's ip in the client logger (cf. https://gitlab.com/front-commerce/front-commerce/issues/23)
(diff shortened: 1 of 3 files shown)


source/docs/reference/configurations.md
```diff
@@ -0,0 +1,107 @@
+---
+id: configurations
+title: Configurations
+---
+
+<blockquote class="note">
+This documentation is the current state of the configurations available in Front-Commerce. However, our goal is to clean it up to make sure that it is still easily customizable, but also easier to check if your configurations are correct. We will provide a clear migration path and will integrate this change within our 1.0.0 release.
+
+Please refer to the [related gitlab issue](https://gitlab.com/front-commerce/front-commerce/issues/106) to follow its advancement.
```
real34:
```suggestion
Please refer to the [related gitlab issue (#106)](https://gitlab.com/front-commerce/front-commerce/issues/106) to follow its advancement.
```

source/docs/reference/configurations.md
```diff
@@ -0,0 +1,107 @@
...
+</blockquote>
+
+## How to set your configurations
+
+Each file described below exist in [`node_modules/front-commerce/src/config`](https://gitlab.com/front-commerce/front-commerce/tree/develop/src/config) in Front-Commerce. If you want to override some, dupplicate those in `my-module/config`.
```
real34:
```suggestion
Each file described below exist in [`node_modules/front-commerce/src/config`](https://gitlab.com/front-commerce/front-commerce/tree/develop/src/config) in Front-Commerce. If you want to override some, duplicate those in `my-module/config`.
```

source/docs/reference/configurations.md
```diff
@@ -0,0 +1,107 @@
...
+
+## `config/website.js`
+
+This configuration file should contain any thing that impacts the content of your website. The term website refers to what a [`website` is in Magento's ecosystem](https://devdocs.magento.com/guides/v2.3/config-guide/multi-site/ms_over.html).
+
+* `root_categories_path` (ex: `1/517/`): which category to use for the main navigation menu. It will then be the children of this category that will be displayed.
+* `default_image_url` (ex: an absolute URL of an image): which image to use when no image path has been given to `<ResizedImage>`
```
real34:
```suggestion
* `default_image_url` (ex: an absolute URL of an image): which image to use when no image path has been given to [`<ResizedImage>`](#TODO)
```

source/docs/reference/configurations.md
```diff
@@ -0,0 +1,107 @@
...
+* `defaultTitle`: the default meta title of your application
+* `defaultDescription`: the default meta description of your application
+* `available_page_sizes`: which page sizes to display in a product list page (or any page with a pagination)
+* `website_id`: which website is used within magento (needed for customer related mutations)
+* `tax` (ex: `1.2` for 20% VAT): Used to correctly filter products within layer queries (this configuration should soon be deprecated with search updates in Front-Commerce)
```
real34:
```suggestion
* `tax` (ex: `1.2` for 20% VAT): Used to correctly filter products within layer queries (this configuration should soon be deprecated with search updates in Front-Commerce — see [#102](https://gitlab.com/front-commerce/front-commerce/issues/102))
```

source/docs/reference/configurations.md
```diff
@@ -0,0 +1,107 @@
...
+* `contentSecurityPolicy`: For security reasons only URLs from the store's domain are authorized through CSPs. However, for tracking and external dependencies, we may authorize more domains. Use the following config and add your custom domain in each :
```
real34:
```suggestion
* `contentSecurityPolicy`: For security reasons only URLs from the store's domain are authorized through CSPs. However, for tracking and external dependencies, we may authorize more domains. Use the following config and add your custom domain in each :<!-- TODO Add a CSP dedicated page -->
```

source/docs/reference/configurations.md
```diff
@@ -0,0 +1,107 @@
...
+title: Configurations
+---
+
+<blockquote class="note">
+This documentation is the current state of the configurations available in Front-Commerce. However, our goal is to clean it up to make sure that it is still easily customizable, but also easier to check if your configurations are correct. We will provide a clear migration path and will integrate this change within our 1.0.0 release.
+
+Please refer to the [related gitlab issue](https://gitlab.com/front-commerce/front-commerce/issues/106) to follow its advancement.
+</blockquote>
+
+## How to set your configurations
```
real34: I think we should a bold warning here, either in introduction or in that most of these configurations are included in the client bundle. Thus it is important to not include private configurations in them, and to use env var instead.

It could also be a good reminder to add in the `website.js` section.

source/docs/reference/environment-variables.md
```diff
@@ -0,0 +1,99 @@
...
+
+These environment variables are the ones available on your server. You can also set them in your `.env` file in your root folder, and they will be loaded on launch.
+
+## Front-Commerce related variables
+
+### Host
+
+Configure the execution environment of the Front-Commerce's application:
+
+- `FRONT_COMMERCE_PORT` (default: 4000): The port of the launched server
```
real34:
```suggestion
- `FRONT_COMMERCE_PORT` (default: `4000`): The port of the launched server
```

source/docs/reference/environment-variables.md
```diff
@@ -0,0 +1,99 @@
...
+- `FRONT_COMMERCE_HOST` (default: 0.0.0.0): The host of the launched server. It might be useful to set it to 127.0.0.1 if you want to only listen local requests.
```
real34:
```suggestion
- `FRONT_COMMERCE_HOST` (default: `0.0.0.0`): The host of the launched server. It might be useful to set it to `127.0.0.1` if you want to only listen local requests.
```

source/docs/reference/environment-variables.md
```diff
@@ -0,0 +1,99 @@
...
+
+- `FRONT_COMMERCE_CACHE_API_TOKEN`: a token that will let external applications invalidate parts of Front-Commerce cache. <!-- TODO link to dataloaders and cache invalidation documentation -->
+
+### Sitemap
+
+- `FRONT_COMMERCE_SITEMAP_TOKEN`: a token that secures the sitemap query in your GraphQL Schema
+
+### PWA
+
+- `FRONT_COMMERCE_DISABLE_OFFLINE_MODE`: in case you don't want to load the offline page when the user is disconnected
```
real34:
```suggestion
- `FRONT_COMMERCE_DISABLE_OFFLINE_MODE`: in case you don't want to load the offline page when the user is offline
```

source/docs/reference/environment-variables.md
```diff
@@ -0,0 +1,99 @@
+---
+id: environment-variables
+title: Environment variables
+---
+
+The environment variables available in Front-Commerce are the configurations that are likely to change depending on the current environment of your application. For instance, you could have three different environments: production, staging and local.
+
+These environment variables are the ones available on your server. You can also set them in your `.env` file in your root folder, and they will be loaded on launch.
```
real34: I think we should make it a bit clearer and make the difference between the build environment and the running server one. Also it would be a good idea to recall that one can access an environment variable under the `process.env` variable.

Finally, it could also be a good think to link to the last section or rewrite here that only `FRONT_COMMERCE_WEB_xxxxxx` variables are available in the React app.

source/docs/reference/configurations.md
JulienPradet:
```suggestion
* `default_image_url` (ex: an absolute URL of an image): which image to use when no image path has been given to `<ResizedImage>` <!-- TODO link to reference -->
```

source/docs/reference/environment-variables.md
```diff
@@ -0,0 +1,124 @@
+---
+id: environment-variables
+title: Environment variables
+---
+
+The environment variables available in Front-Commerce are the configurations that are likely to change depending on the current environment of your application. For instance, you could have three different environments: production, staging and local.
+
+These environment variables can be defined in two diferrent ways:
```
real34:
```suggestion
These environment variables can be defined in two different ways:
```

source/docs/reference/environment-variables.md
```diff
@@ -0,0 +1,124 @@
...
+* on your server (See [How To Read and Set Environmental and Shell Variables on a Linux VPS](https://www.digitalocean.com/community/tutorials/how-to-read-and-set-environmental-and-shell-variables-on-a-linux-vps))
+* in the `.env` file in your root folder
+
+You can then access them by using the `process.env` object in your javascript files. This variable is available in both your server side code and client side code. However,
```
real34: It could be misleading… here is a proposal:

```suggestion
You can then access them by using the `process.env` object in your javascript files no matter if it is a server-side or client-side file. However,
```

source/docs/reference/environment-variables.md
```diff
@@ -0,0 +1,124 @@
...
+not all variables are exposed in your client code. Client code only have access to variables such as `FRONT_COMMERCE_WEB_*`. See [Add your own environment variables](http://localhost:4444/docs/reference/environment-variables.html#Add-your-own-environment-variables) for more details.
+
+## How to update environment variables
+
+However, in some cases you can't update these variables only by updating your server's variable.
```
real34: To discuss.

JulienPradet: WIP

PR #59, opened 2019-03-28
Add a note about root Query type limitation

real34:
in schema stitching

See https://gitlab.com/front-commerce/front-commerce/issues/196 for more context

### DIFF
--- a/source/docs/advanced/graphql/remote-schemas.md
+++ b/source/docs/advanced/graphql/remote-schemas.md
@@ -182,8 +182,28 @@ module.exports = {
 
 ## Caveats
 
+### Server restart needed
+
 Please note that **remote schema changes will only appear in your schema after a server restart**.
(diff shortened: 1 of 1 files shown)


source/docs/advanced/graphql/remote-schemas.md
```diff
@@ -182,9 +182,29 @@ module.exports = {
...
 ## Caveats
 
+### Server restart needed
+
 Please note that **remote schema changes will only appear in your schema after a server restart**.
 The remote schema definition is fetched during the server bootstrap process.
 
+### Query root type must be provided
+
+The remote schema (after transforms applied) **MUST** contain at least a `Query` root type.
```
JulienPradet: I would mention that this is the case when you only define a mutation or that you filter all the Queries from your remote schema.

PR #74, opened 2019-05-16
Add minimal documentation for WYSIWYG components

JulienPradet:


source/docs/reference/wysiwyg.md
```diff
@@ -0,0 +1,45 @@
...
+
+For a tangible example and explanation, please refer to [Display WYSIWYG content](/docs/advanced/theme/wysiwyg.html) advanced documentation.
+
+## `EnhanceWysiwyg`
+
+### Parameters
+
+* `option`: the first parameter is an object with the following properties
+    * `richTextPropName` string (default: "content"): the name of the new property which contains the `html` when using the enhanced component.
+    * `shortcodes` function: An function that returns the list of shortcodes used in the WYSIWYG. Each shortcode is an object which contains the keys:
```
real34:
```suggestion
    * `shortcodes` function: a function that returns the list of shortcodes used in the WYSIWYG. Each shortcode is an object which contains the keys:
```

source/docs/reference/wysiwyg.md
```diff
@@ -0,0 +1,45 @@
...
+        * `regex` RegExp: the regex that allows to match the shortcode within the HTML
+        * `replacement` (second argument of [String.prototype.replace](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace)): how should the matched string should be replaced
```
real34:
```suggestion
        * `replacement` (second argument of [String.prototype.replace](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace)): how should the matched string be replaced
```

source/docs/reference/wysiwyg.md
```diff
@@ -0,0 +1,45 @@
...
+    * `transforms` array: the list of transforms that should be applied to your HTML. Each transform is a function with the following signature `(node, convertNodeToElement) => null|ReactElement` where:
+        * `node` is a parsed node element as defined in [react-html-parser](https://www.npmjs.com/package/react-html-parser#arguments-1)
+        * `convertNodeToElement` allows to parse children of the current `node`
+            Usually used like this `node.children.map(convertNodeToElement)`
+        * the result is either null if the node does not match the current transform or a React Element if it matches.
```
real34:
```suggestion
        * the result is either `null` if the node does not match the current transform or a [React Element](https://reactjs.org/docs/rendering-elements.html) if it matches.
```

PR #76, opened 2019-05-27
Add a note about the new `FRONT_COMMERCE_UNSAFE_INSECURE_MODE` env var

real34:


JulienPradet:
Looks good.

What concerns me is what value should one set in `FRONT_COMMERCE_UNSAFE_INSECURE_MODE`? (cf. my other comment on gitlab)

PR #80, opened 2019-06-05
Submit payment

JulienPradet:
Here is a very small migration guide for payments updates.

People needing to upgrade should contact us rather than trying this on their own.

### DIFF
--- a/source/docs/appendices/migration-guides.md
+++ b/source/docs/appendices/migration-guides.md
@@ -24,6 +24,16 @@ If you experience issues after the upgrade, here are the things to ensure:
 
 If you still experience issues, please [contact us](mailto:support@front-commerce.com).
 
+### Embedded Payments
+
+Previously, when a payment needed to set custom information, there were two steps: one for validating the payment information, and one for placing the order.
+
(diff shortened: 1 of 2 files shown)


source/docs/appendices/migration-guides.md
```diff
@@ -24,6 +24,16 @@ If you experience issues after the upgrade, here are the things to ensure:
 
 If you still experience issues, please [contact us](mailto:support@front-commerce.com).
 
+### Embedded Payments
+
+Previously, when a payment needed to set custom information, there were two steps: one for validating the payment information, and one for placing the order.
+
+This is no longer the case, but this means that AddtionalPaymentInformation components need to use the new behavior, which is to use `theme/modules/Checkout/Payment/SubmitPayment` when validating the additional payment information.
```
real34:
```suggestion
This is no longer the case, which improves the Customer experience. But this means that AddtionalPaymentInformation components need to use the new behavior, which is to use `theme/modules/Checkout/Payment/SubmitPayment` when validating the additional payment information.
```

PR #104, opened 2019-12-18
Add documentation pages about preloading and FRONT_COMMERCE_FAST

JulienPradet:


source/docs/advanced/performance/preloading-routes.md
```diff
@@ -0,0 +1,183 @@
+---
+id: preloading-routes
+title: Preloading Routes' data
+---
+
+When a website is slow, the first thing that comes to mind is to improve raw metrics like the Speed Index. And while this is definitly useful, we shouldn't overlook the [_perceived_ performances](https://blog.teamtreehouse.com/perceived-performance). The goal is not to improve the response time, but to better manage the time to make the user _think_ that the website is fast.
+
+One way to do this is to anticipate the user's need and start the long running tasks before they actually need it. For instance, when a user is navigating on a catalog, on desktop, they will hover the product before clicking on it. Thus, we can trigger the product's load as soon as they hover the product, instead of waiting for the click. This can seem like small gains, but this definitly improves how fast the user _thinks_ your site is.
```
real34:
```suggestion
One way to do this is to anticipate the user's need and start the long running tasks before they actually need it. For instance, when a user is navigating on a catalog, on desktop, they will hover the product before clicking on it. Thus, we can start loading the product page as soon as the user hover the product element, instead of waiting for the click. This can seem like a micro-optimization, but it definitely improves how fast the user _thinks_ your site is.
```

source/docs/advanced/performance/preloading-routes.md
```diff
@@ -0,0 +1,183 @@
...
+
+## How to preload data?
+
+The main idea is to launch Apollo requests twice:
+
+- when you think that the user will need the data
+- when you are displaying the data
+
+By doing this, you will first warm up Apollo's cache and this cache will be reused when you will display your data. But this can be tedious. This is why in Front-Commerce, we have added a <abbr title="Higher Order Component">HOC</abbr> that will do this: `web/core/apollo/graphqlWithPreload` (added in version 1.0.0-rc.0).
```
real34:
```suggestion
By doing this, you will first warm up Apollo's cache and this cache will be reused when you will display your data. But this can be tedious. This is why in Front-Commerce, we have added a <abbr title="Higher Order Component">HOC</abbr> that will do this: `web/core/apollo/graphqlWithPreload`.
```

source/docs/advanced/performance/preloading-routes.md
```diff
@@ -0,0 +1,183 @@
+---
+id: preloading-routes
+title: Preloading Routes' data
+---
+
```
real34: I think we should add this here for the whole page for now. At some point we could remove all those notices when we will be confident that all projects use 1.0 for instance (and later, after EOL).

```suggestion

<blockquote class="feature--new">
  _This feature has been added in version `1.0.0-rc.0`_
</blockquote>

```

source/docs/advanced/performance/preloading-routes.md
```diff
@@ -0,0 +1,183 @@
...
++    )
++  )
+  graphqlWithPreload(Query)
+)(BaseComponent)
+  ```
+</blockquote>
+
+## How to preload routes?
+
+However, in some cases you don't really know which component you should preload. This is the case for the url: `/venia-dresses` for instance. If you are familiar with Venia's sample data, you will recognize that it is a category. But you may not know which component is actually used to display it.
```
real34:
```suggestion
However, in some cases you don't really know which component you should preload. This is the case for the url: `/venia-dresses` for instance. If you are familiar with Magento Venia's sample data, you will recognize that it is a category. But you may not know which component is actually used to display it.
```

source/docs/advanced/performance/preloading-routes.md
```diff
@@ -0,0 +1,183 @@
...
+
+This is why we have implemented preloads on links too. This is possible with the React hook `theme/modules/Router/usePreload` (added in version 1.0.0-rc.0) that will give you a preload function to load the components and data for a given link.
```
real34:
```suggestion
This is why we have implemented preloads on links too. This is possible with the React hook `theme/modules/Router/usePreload` that will give you a preload function to load the components and data for a given link.
```

source/docs/advanced/performance/preloading-routes.md
```diff
@@ -0,0 +1,183 @@
...
+
+If you need search parameters in the URL (e.g. `?param=value`), it will be available at `params.location.search`.
+
+## Front-Commerce Fast Mode
+
+To conclude, if you want your application to feel even faster to your users, you need to:
+* add a `preload` static property to your routes thanks to `graphqlWithPreload`
+* use `usePreload` hook through the default Link component in Front-Commerce or manually in your own events.
+
+This also enables a wider range of optimizations such as Front-Commerce Fast Mode. Please refer to [FRONT_COMMERCE_FAST] for more information.
```
real34: The `FRONT_COMMERCE_FAST` link is missing here.

source/docs/advanced/performance/faster-server-side-rendering.md
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+}
+```
+
+## Limit the number of render iteration
+
+But there is also in Front-Commerce the possibility to force only one iteration of the loop mentioned above. This can be enabled by setting the `FRONT_COMMERCE_FAST` environment variable to `true`.
+
+This can improve your SSR times drastically. However the risk is that you may have some data that are not rendered properly at the top level of your page. This is a tradeoff between safety and speed.
```
real34: I'm confused about this sentence.

Here is a proposal:
```suggestion
This can improve your SSR times drastically. However you must be aware that only the top-level queries will be executed on the server. The generated HTML may not contain data that are important for your users, or Search Engines. Such content would be loaded client-side though. This is a tradeoff between safety and speed.
```

source/docs/advanced/performance/preloading-routes.md
JulienPradet:
```suggestion
This also enables a wider range of optimizations such as Front-Commerce Fast Mode. Please refer to [FRONT_COMMERCE_FAST](/docs/advanced/performance/faster-server-side-rendering.html) for more information.
```

PR #108, opened 2020-02-12
Reworked the public roadmap to be about our release process

real34:
since we saw that our roadmap was moving fast due to customers feedbacks


source/docs/appendices/release-process.md
```diff
@@ -0,0 +1,20 @@
+---
+id: release-process
+title: Release process
+---
+
+Front-Commerce is built around the motto "Make it work, then make it better". This means that it is possible **today** to develop and deploy your website to production. But we still have a lot of things to work on to accomplish our mission: make your life easier, and your customers happier.
```
JulienPradet: The "But we still have a lot of things to work on" feels a bit negative. What's in FC today is already great for external developers. This phrasing is a bit downsizing what we can provide.

source/index.md
```diff
@@ -111,8 +111,8 @@ layout: index
   </ul>
 
   <p>
-    <a class="link learn-more button" href="/docs/appendices/roadmap.html">
-      Take a look at our Roadmap <span class="icon-arrow-right-alt"></span>
+    <a class="link learn-more button" href="mailto:contact@front-commerce.com?subject=Hi!%20I'd%20like%20to%20know%20more%20about%20your%20roadmap&body=Hi%2C%0D%0A%0D%0AI%20am%20interested%20in%20your%20product%2C%20and%20would%20love%20to%20know%20if%20you%20considered%20...">
```
JulienPradet: Since that's no longer a link but more of a CTA, maybe it should be a button like link?

PR #113, opened 2020-02-27
Caching

real34:


source/docs/advanced/graphql/dataloaders-and-cache-invalidation.md
```diff
@@ -0,0 +1,396 @@
...
+    }
+  ]
+};
+```
+
+### Advanced usage
+
+If you need additional implementations or want to leverage strategies for a specific use case, please [contact us](mailto:contact@front-commerce.com) so we can discuss it and guide you!
+
+## Invalidating the cache
```
JulienPradet: Not sure when it'll make sense to document it, but it might prove to be useful to expose/document the node API to invalidate cache directly. I'm thinking of systems that only have webhooks (SaaS platforms mainly) where we can add a server endpoint that will then call the correct invalidation method in node.

source/docs/reference/configurations.md
```diff
@@ -130,25 +130,44 @@ module.exports = ["products", "categories", "pages"];
...
 
-<blockquote class="wip">
-**Work In Progress:** more details about specific configurations will be detailed as part of https://github.com/front-commerce/developers.front-commerce.com/issues/49. If you need it right away, please [contact us](mailto:contact@front-commerce.com). We will make sure to answer you in a timely manner.
+* `defaultMaxBatchSize`: default batch size used for dataloaders (unless specified during instantiation) (default: `100`)
+* `strategies`: list of caching strategies to use in the application (default: `[]`)
+
+Each strategy can be configured with the keys below:
+
+* `implementation`: name of the [implementation of this strategy](/docs/advanced/graphql/dataloaders-and-cache-invalidation.html#Caching-strategies) (**mandatory**)
+* `supports`: list of loaders impacted by this strategy. Either an array of values or `"*"` for all. (default: `[]`)
```
JulienPradet: Wouldn't `*` be a better default here? Or should it be mandatory? Setting up a strategy that does nothing does not seem useful.

real34: That is something I asked myself during the implementation, but I was even more afraid by the impacts of adding a caching strategy to **ALL** dataLoaders by mistake (e.g a typo with `support` instead of `supports`).

Maybe we should make this a mandatory parameter instead, as I agree that it makes no sense to have something that does nothing by default.

JulienPradet: Let's take the mandatory route then. :)

source/docs/advanced/graphql/dataloaders-and-cache-invalidation.md
```diff
@@ -0,0 +1,396 @@
+---
+id: dataloaders-and-cache-invalidation
+title: DataLoaders and cache invalidation
+---
+
+One of the main responsibility of GraphQL modules in Front-Commerce is fetching data from remote sources in resolvers. A naive approach may reach a few limits in a real application, that is why **Front-Commerce provides helpers allowing developers to build high-performance applications**.
+
+This section introduce the DataLoader pattern and how to use it in a Front-Commerce application. It then details how to configure the caching layer to keep remote API response in cache to boost performance. Finally, it explains how to invalidate the cache from remote systems.
+
+By the end of this reading, you will have a good understanding of the mechanisms at your disposal to compensate for the slowness of a remote API.
```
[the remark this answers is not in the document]
real34: It seems (https://www.deepl.com/fr/translator#en/fr/By%20the%20end%20of%20this%20reading%2C%20you%20will%20have%20a%20good%20understanding%20of%20the%20mechanisms%20at%20your%20disposal%20to%20compensate%20for%20the%20slowness%20of%20a%20remote%20API.%0A) but I'll use "guide"

source/docs/advanced/graphql/dataloaders-and-cache-invalidation.md
```diff
@@ -0,0 +1,396 @@
...
+      name
+      qty
+    }
+  }
+}
+```
+
+### The problem
+
+This query would lead to 11 HTTP requests from the server to the remote datasource:
```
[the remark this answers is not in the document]
real34: Well spotted! Thanks, I'll add a `name` on `Category` and use 12.

source/docs/advanced/graphql/dataloaders-and-cache-invalidation.md
```diff
@@ -0,0 +1,396 @@
...
+
+If the product "PANT-01" was an upsell of all other products, the inventory API would be requested again when resolving the `upsells.qty` field. Query-level caching prevents such extra-calls, by reusing the response from `products.qty` that has already been fetched previously to resolve data.
+
+## What are DataLoaders?
+
+DataLoader is a pattern promoted by Facebook, from their internal implementations, to solve problems with data fetching. We use this name because it is the name of the [reference implementation in Javascript: graphql/dataloader](https://github.com/graphql/dataloader).
+
+A dataLoader is instantiated with a **batching function**, that will allow to fetch data in a grouped way (see [Batching](#Batching) above). It also has a caching strategy that prevents fetching the same data twice in the same request or across requests (see [Caching](#Caching) above).
+
+By default every dataLoader provides request-level caching. Caching configuration allow developers to use a persistent/in-memory caching strategy such as Redis to share the cache between users and requests.
```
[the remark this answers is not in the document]
real34: It is not clear indeed. In-memory was for "in RAM".
I'll remove it.

source/docs/advanced/graphql/dataloaders-and-cache-invalidation.md
```diff
@@ -0,0 +1,396 @@
...
+}
+```
+
+If the product "PANT-01" was an upsell of all other products, the inventory API would be requested again when resolving the `upsells.qty` field. Query-level caching prevents such extra-calls, by reusing the response from `products.qty` that has already been fetched previously to resolve data.
+
+## What are DataLoaders?
+
+DataLoader is a pattern promoted by Facebook, from their internal implementations, to solve problems with data fetching. We use this name because it is the name of the [reference implementation in Javascript: graphql/dataloader](https://github.com/graphql/dataloader).
+
+A dataLoader is instantiated with a **batching function**, that will allow to fetch data in a grouped way (see [Batching](#Batching) above). It also has a caching strategy that prevents fetching the same data twice in the same request or across requests (see [Caching](#Caching) above).
```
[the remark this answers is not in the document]
real34: It indeed seems to have changed. I'll update it.

source/docs/advanced/graphql/dataloaders-and-cache-invalidation.md
```diff
@@ -0,0 +1,396 @@
...
+    };
+  }
+};
+```
+
+The 2nd parameter to `makeDataLoader` are the options to pass to the dataLoader instance. You usually don't have to use it but if you do, please refer to [dataloader's documentation](https://github.com/graphql/dataloader#new-dataloaderbatchloadfn--options) for further information.
+
+### Useful patterns
+
+#### Prevent caching not found data
```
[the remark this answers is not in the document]
real34: Let's give a try to "Prevent caching errors (data not found)"

source/docs/advanced/graphql/dataloaders-and-cache-invalidation.md
real34: I suggest that we wait for a few implementations of this pattern to go live before documenting it. I feel like some abstraction may emerge.

source/docs/reference/configurations.md
real34: Done in https://gitlab.com/front-commerce/front-commerce/-/merge_requests/188.
Documentation updated by anticipation.

PR #114, opened 2020-02-27
Document SSR and related mechanisms in place

real34:
There is a link to the yet to be finalized Configuration page, we must be careful about not breaking it when merging the other PR.

### DIFF
--- a/source/docs/advanced/theme/server-side-rendering.md
+++ b/source/docs/advanced/theme/server-side-rendering.md
@@ -61,6 +61,10 @@ export default branchServerClient(
 )(MyClientSideComponent);
 ```
 
+<blockquote class="tip">
+  **Tip:** the function name (branch**Server**_Client_) is a good way to remember that the first parameter is what is rendered on the **server**, and the second on the _client_.
+</blockquote>
+
 If you only need to know the rendering context, you can use Front-Commerce's `withIsServer` HOC to have an `isServer` prop injected in your component.
 
 Example:
@@ -109,7 +113,7 @@ If you want to tweak this behavior, we invite you to browse the [`deviceConfigPr
   _This feature has been added in version `2.0.0-rc.0`_
 </blockquote>
 
-During development you might face some SSR errors due to syntax errors or [browser API usage](/docs/advanced/theme/server-side-rendering.html#There-is-no-window-on-the-server). By default in dev mode, Front-Commerce provides an error page that makes the mistake obvious to allow you to fix it as early as possible.
+During SSR you might face some errors (syntax errors, [browser API usage](/docs/advanced/theme/server-side-rendering.html#There-is-no-window-on-the-server), render errors, etc.). By default in dev mode, Front-Commerce provides an error page that makes the mistake obvious to allow you to fix it as early as possible.
 
 Hopefully the previous screen will prevent such errors to occur in production, but development is not an exact science and things can go wrong! In such situations, the `theme/pages/SsrFallback` component will be rendered instead of the page. **It displays a clean loading area, but you can override it to provide another visual waiting page to users while the application is being rendered client side.**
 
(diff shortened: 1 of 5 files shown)


source/docs/advanced/theme/server-side-rendering.md
```diff
@@ -0,0 +1,120 @@
+---
+id: server-side-rendering
+title: Server Side Rendering (SSR)
+---
+
+Front-Commerce uses <abbr title="Server Side Rendering">SSR</abbr> (opposed to <abbr title="Client Side Rendering">CSR</abbr>) to improve <abbr title="Search Engine Optimization">SEO</abbr> and <abbr title="User experience">UX</abbr> by serving <abbr title="Hypertext Markup Language">HTML</abbr> pages to users on their first hit to the server. You can read [A typical request in Front-Commerce](/docs/concepts/architecture-overview.html#A-typical-request-in-Front-Commerce) to get a better understanding of this process.
```
JulienPradet: Acronym party! 🤸

source/docs/advanced/theme/server-side-rendering.md
```diff
@@ -0,0 +1,120 @@
...
+```js
+import branchServerClient from "web/core/branchServerClient";
+
+const MyClientSideComponent => props => <div>Hi!</div>;
+
+export default branchServerClient(
+  () => () => null, // do not render anything during SSR
+  BaseComponent => BaseComponent // render the base component on the client
+)(MyClientSideComponent);
+```
```
JulienPradet: Not sure if it's worth mentioning this tip here but I know it helped some of our integrators:

To memorize which element renders on the server and which one renders on the client, refer to the method name: `branchServerClient`, thus the first parameter is for the server and the second one is for the client.

source/docs/advanced/theme/server-side-rendering.md
```diff
@@ -0,0 +1,120 @@
...
+  My latest render occured on the {props.isServer ? "server" : "client" }.
+</div>;
+
+export default withIsServer()(MyIsomorphicComponent);
+```
+
+<blockquote class="note">
+  **ProTip™:** in order to reduce the amount of Javascript sent to users, we've found that code using `branchServerClient()` are often good candidates to code splitting. You may consider using a `loadable()` component here. You can use [`DEBUG="front-commerce:webpack"`](/docs/reference/environment-variables.html#Debugging) to analyze your bundles and decide wether it would improve your application or not.
+</blockquote>
+
```
JulienPradet: Should we mention the fact that the server part will always be rendered first before rendering the client? This stands even if the user is navigating client side.

source/docs/advanced/theme/server-side-rendering.md
```diff
@@ -0,0 +1,120 @@
...
+If you want to tweak this behavior, we invite you to browse the [`deviceConfigProvider`](https://gitlab.com/front-commerce/front-commerce/-/blob/master/src/server/express/ssr/deviceConfigProvider.js) for implementation details and [contact us](mailto:contact@front-commerce.com) if you think of improvements.
+
+## SSR Fallback when things go wrong
+
+<blockquote class="feature--new">
+  _This feature has been added in version `2.0.0-rc.0`_
+</blockquote>
+
+During development you might face some SSR errors due to syntax errors or [browser API usage](/docs/advanced/theme/server-side-rendering.html#There-is-no-window-on-the-server). By default in dev mode, Front-Commerce provides an error page that makes the mistake obvious to allow you to fix it as early as possible.
```
JulienPradet:
```suggestion
During SSR you might face some errors (syntax errors, [browser API usage](/docs/advanced/theme/server-side-rendering.html#There-is-no-window-on-the-server), render errors, etc.). By default in dev mode, Front-Commerce provides an error page that makes the mistake obvious to allow you to fix it as early as possible.
```

source/docs/advanced/theme/server-side-rendering.md
real34: What are the pitfalls of this? I am not sure how to mention it here without making the whole page too cumbersome.

I'd appreciate a hand for this, to see how it could fit in the big picture. I suggest to add it in an upcoming PR after merging this one.

PR #116, opened 2020-02-27
Explain how configuration providers work

JulienPradet:


source/docs/advanced/server/configurations.md
```diff
@@ -0,0 +1,246 @@
+---
+id: setup-configurations
+title: Setup configurations
```
real34: I am not sure that people would understand from the title what this section is about (maybe because it is not my mother-tongue). I don't have any final suggestion for it though.

Maybe "Configurations" would be sufficient.

Or
```suggestion
title: Configure server modules
```

source/docs/advanced/server/configurations.md
```diff
@@ -0,0 +1,246 @@
...
+---
+
+Configurations is a part of software development that can quickly become messy. Validation, detection, performance... There is a lot to take care of. In Front-Commerce we solved this by defining what we call **Configuration Providers**.
```
real34:
```suggestion
Configuration is a part of software development that can quickly become messy. Validation, detection, performance... there is a lot to take care of. In Front-Commerce we solved this by defining what we call **Configuration Providers**.
```

source/docs/advanced/server/configurations.md
```diff
@@ -0,0 +1,246 @@
...
+
+In this guide we will see how it works and how to create your own to override existing configurations or create new ones.
```
real34:
```suggestion
In this guide we will see how they work and how to create your own create new configurations or override existing ones.
```

source/docs/advanced/server/configurations.md
```diff
@@ -0,0 +1,246 @@
...
+
+## What is a configuration provider?
+
+The goal is to give access to a configuration object that contains all the configuration of the store. This configuration object can contain any kind of configuration. It can be flags about a feature activation, some credentials to connect to a remote service, etc.
+
+However, not all applications will need every single configuration. For instance, a shop that chose to use [Algolia](https://www.algolia.com/) won't have the same configurations than a shop that chose [Elasticsearch](https://www.elastic.co/) for product search. The goal of the configuration providers is to define the configurations needed for the specific modules you use in your application.
+
+On server start, the configuration providers will then be fused to create the global configuration object. This configuration object will then available on each server request.
```
real34:
```suggestion
On server start, the configuration providers will then be combined to create the global configuration object. This configuration object will then available on each server request.
```

source/docs/advanced/server/configurations.md
```diff
@@ -0,0 +1,246 @@
...
+    "proxiedPaths": []
+  }
+}
+```
+
+> **Important:** Don't activate this on a public instance as it can expose sensitive data
+
+## Configuration provider definition
+
+In practice, a configuration provider is an object that have four parts: a name, a schema, static values and values depending on the request.
```
real34:
```suggestion
In practice, a configuration provider is an object that have four parts: a name, a schema, static values (available independently from the request) and values depending on the request.
```

source/docs/advanced/server/configurations.md
```diff
@@ -0,0 +1,246 @@
...
+      env: "FRONT_COMMERCE_SERVICE_SECRET"
+    }
+  }
+};
+```
+
+You could then access the values with `req.config.service.key` or `req.config.service.secret`.
+
+Please note that if another configuration provider's schema also defined a `service` key, it would merge the definitions and in your final `req.config.service` you would have both the keys from the other configuration provider's schema and from the above schema.
+
```
real34: Should we mention (in a blockquote for instance) the `url-without-slash` and `collection` types  available in Front-Commerce?

source/docs/advanced/server/configurations.md
```diff
@@ -0,0 +1,246 @@
...
+### Static values (key `values`, optional)
+
+Most of the time `default` values in the schema is sufficient. However, in some cases you may need to fetch some values from an API, a dynamic file, etc. This is what `values` is for.
+
+It is an optional promise that should return the missing values in your schema. It will override default values and env values from the schema with the new values. However if in your `values` result there is some keys that are missing the default values and env values will be kept. For instance, if we implemented the following `values` promise, the `secret` would still be `process.env.FRONT_COMMERCE_SERVICE_SECRET` from the above schema.
```
real34: I am not convinced by my suggestion, but think the initial sentence is a bit difficult to read.
```suggestion
It is an optional promise that should return the missing values in your schema. It will override default values and env values from the schema with the new values. However only keys from the `values` result will take precedence, the default values and env values will be kept for other keys. For instance, if we implemented the following `values` promise, the `secret` would still be `process.env.FRONT_COMMERCE_SERVICE_SECRET` from the above schema.
```

source/docs/advanced/server/configurations.md
```diff
@@ -0,0 +1,246 @@
...
+  }));
+```
+
+Please note that this promise is launched only once on server bootstrap. If the configuration changes over time on your API, the `req.config.service.key` value will still be the same.
+
+If it needs to change over time, please use `slowValuesOnEachRequest` instead.
+
+### Values depending on the request (key `slowValuesOnEachRequest`, optional)
+
+> **Note:** `slowValuesOnEachRequest` is a low level API in Front-Commerce. You shouldn't need it in most cases
```
real34:
```suggestion
> **Note:** `slowValuesOnEachRequest` is a low level API in Front-Commerce. You shouldn't need it in most cases.
```

source/docs/advanced/server/configurations.md
```diff
@@ -0,0 +1,246 @@
...
+
+`slowValuesOnEachRequest` is a function that extracts configuration values from the current request. For instance, depending on the URL, the configuration `currentShopId` will get a different id and thus display different informations.
```
real34:
```suggestion
`slowValuesOnEachRequest` is a function that extracts configuration values from the current request. For instance, depending on the URL, the configuration `currentShopId` will get a different id and thus display different information.
```

source/docs/advanced/server/configurations.md
JulienPradet: I don't think it's relevant for now.

PR #117, opened 2020-03-02
Document the GraphQL schema change of 2.0.0-rc.0

JulienPradet:


source/docs/appendices/migration-guides.md
```diff
@@ -190,6 +190,33 @@ For more information, please have a look at the [Routes reference](/docs/referen
     * If it's a file you've overridden from Front-Commerce's core, please check in the core how the file changed. If it's still used, the `<Route>` components have most likely been replaced with the children property. If it's not, it usually means that it is now replaced by a layout. If you're not sure, feel free to contact our team. We will make sure to make this as painless as possible.
 </details>
 
+## ID types in GraphQL definitions
+
+In order to improve consistency of the GraphQL schema accross platforms, we have also made some changes to the GraphQL schema.
+
+All ids now leverage the `ID` type instead of being `Int` or `String`. This will allow better compatibility across platforms since depending on the platform it can either be an `Int` or a `String` but shouldn't affect the final behavior of your components. If you had some in your own schemas, we suggest to also convert them to `ID`.
```
real34:
```suggestion
All ids now leverage the `ID` type instead of being `Int` or `String`. This will allow better compatibility across platforms since `ID` allow for `Int` or `String` values as identifiers. It shouldn't affect the final behavior of your components. If you had some in your own schemas, we suggest to also convert them to `ID`.
```

source/docs/appendices/migration-guides.md
```diff
@@ -190,6 +190,33 @@ For more information, please have a look at the [Routes reference](/docs/referen
...
+
+The only thing that you need to make sure of is to run `npm run lint` on your project. This should allow you to see errors on overridden GraphQL files.
```
real34: I think we should warn too about `===` and input parameters (cast values). For instance, url params may have been casted to `Number`.

source/docs/appendices/migration-guides.md
```diff
@@ -190,6 +190,33 @@ For more information, please have a look at the [Routes reference](/docs/referen
     * If it's a file you've overridden from Front-Commerce's core, please check in the core how the file changed. If it's still used, the `<Route>` components have most likely been replaced with the children property. If it's not, it usually means that it is now replaced by a layout. If you're not sure, feel free to contact our team. We will make sure to make this as painless as possible.
 </details>
 
+## ID types in GraphQL definitions
```
real34:
```suggestion
### ID types in GraphQL definitions
```

source/docs/appendices/migration-guides.md
JulienPradet:
```suggestion
The only thing that you need to make sure of is to run `npm run lint` on your project. This should allow you to see errors on overridden GraphQL files.

Please note that if in your code, you relied on some int types, there might be cases where identifiers comparison won't work. All code within the core's module has been update to avoid as much compatibility issues as possible. However keep in mind that if you rely on some strict equality (`===`) and static identifiers, this could lead to issues.
Please note that if in your code, you relied on some int types, there might be cases where they won't be cast to integers anymore. All code within the core's module has been update to avoid as much compatibility issues as possible. However keep in mind that if you rely on some strict equality (`===`) and static identifiers, this could lead to issues.
```

What do you think about this?

real34: Let's go with it!

Minor typo:
> has been updated

PR #119, opened 2020-03-05
Magento1 doc - installation steps update // Expose attributes

bastienlm:


source/docs/magento1/exposing-additional-attributes.md
```diff
@@ -0,0 +1,106 @@
+---
+id: exposing-additional-attributes
+title: Exposing additional attributes
+---
+
+In case where you need to expose new attributes on the existing endpoint you can simply add this attribute on your `api2.xml` file. For that, you need to retrieve resource name (API_UNIQUE_NAME value if you follow the basic structure shown in "Add custom endpoint section") and add your attribute(s) into the `attributes` node.
```
JulienPradet: Before starting the documentation page, we should explain why this can be useful. A message like:

> In many shops, there are more attributes than the standard ones available in Magento. This can be done following this [Magento documentation] but it then needs to be fetched by Front-Commerce. This can be done by exposing the newly created attributes in the endpoints already used in Front-Commerce. For instance, when retrieving products, Front-Commerce uses the `/api/rest/products` URL.
>
> In this documentation guide, you will learn how to add these attributes to any endpoint.


Then you can introduce the api2.xml, and continue with your documentation.

source/docs/magento1/exposing-additional-attributes.md
```diff
@@ -0,0 +1,106 @@
...
+
+- How to retrieve the resource name: you can retrieve this data on all `api2.xml` file.
+- Expose attribute `custom_attribute` for customer data:
```
JulienPradet: I wouldn't use this listing presentation and prefer some categories.

For instance here it's a section about adding `custom_attribute` for the most common resources.

source/docs/magento1/exposing-additional-attributes.md
```diff
@@ -0,0 +1,106 @@
...
+    <category>
+        <attributes>
+            <custom_attribute />
+        </attributes>
+    </customers>
+</resources>
+...
+```
+- Forcing attributes:
+Sometime some attributes is forcing exclude. You can retrieve an example on `app/code/community/Clockworkgeek/Extrarestful/etc/api2.xml` file:
```
JulienPradet: I don't understand what you mean here. Is it about excluding an attribute or allowing to retrieve an attribute?

source/docs/magento1/exposing-additional-attributes.md
JulienPradet: Even though there is another page explaining where to put the `api2.xml`, we should repeat it here.

source/docs/magento1/exposing-additional-attributes.md
```diff
@@ -0,0 +1,106 @@
...
+                <image>1</image>
+                <is_active>1</is_active>
+            </read>
+        </guest>
+    </exclude_attributes>
+    ...
+</category>
+...
+```
+for avoid this exclude, you can simply retrieve node who you want to avoid in your own `api2.xml` file and set to `0`.
```
JulienPradet: Set what to `0` ?

> If you want to include the attribute instead, you can retrieve...

source/docs/magento1/exposing-additional-attributes.md
```diff
@@ -0,0 +1,106 @@
...
+
+example for avoiding image restriction for guest and customer:
```
JulienPradet: There should be a leading capital letter.

> for avoiding -> to avoid 

What's the image restriction? Is it by default? It would make sense to write explicitly what kind of image are you talking about in this example.

real34:
Hi!

Thanks for your contributions.\
I applied @JulienPradet's recommendations in  7ad69c13f3112a8e5206860f1c901ede54523d10.

PR #122, opened 2020-03-16
Update doc to use `export default` where relevant

real34:
I may be wrong as I did a search/replace on the whole doc by filtering some files where a `module.exports` is needed.


real34:
Closing this as part of cleaning because it's old and may have to be done later again.

PR #126, opened 2020-06-03
Add explanation of WysiwygV2 component

JulienPradet:
This is a *lot* of information at once. Not sure how we should simplify it.

Still marked as WIP because I'd like to take this night back to reread it with a fresher eye tomorrow morning.


source/docs/advanced/theme/wysiwyg.md
```diff
@@ -4,116 +4,385 @@ title: Display WYSIWYG content
...
+You can now transform HTML tags into React components. However, in some cases the data provided in the raw HTML you've received from your remote service won't be enough to display your component. For instance, you may have a `<product-name sku="VSK12" />` that is supposed to display the product's name. But all you have in the attributes is the SKU. This can be done by associating a `WysiwygNodeData` with your HTML node.
 
-For instance, we want to make sure that it transforms `<a>` tags to `Link` elements to preserve React's navigation. For instance, `<a href="/about-us">About us</a>` will be transformed to `<Link to="/about-us">About us</Link>`. It looks like this:
+Indeed, when you've created your `CustomWysiwyg` type, you have declared: `childNodes` (`JSON`) and `data` (`[WysiwygNodeData]`). By default, `data` will always be empty. But you can tell the Wysiwyg loader to fetch data for some specific `childNodes` (that each represent an HTML tag). This means that for `product-name`, you will associate a new `data` element that will contain the whole product data needed to display the component. To do so, you need to follow these steps:
```
real34, line 167: The transition is misleading as readers may think that these types are about client-side code.
```suggestion
Indeed, when you've created your GraphQL `CustomWysiwyg` type, you have declared: `childNodes` (`JSON`) and `data` (`[WysiwygNodeData]`). By default, `data` will always be empty. But you can tell the Wysiwyg loader to fetch data for some specific `childNodes` (that each represent an HTML tag). This means that for `product-name`, you will associate a new `data` element that will contain the whole product data needed to display the component. To do so, you need to follow these steps:
```

Ideally, we could also transform "you've created your GraphQL `CustomWysiwyg` type" into a link to the correct anchor.

source/docs/advanced/theme/wysiwyg.md
```diff
@@ -4,116 +4,385 @@ title: Display WYSIWYG content
...
 ```
+    Feel free to split this in a Fragment dedicated to your `CustomWysiwyg` just like it is done in the core for `MagentoWysiwyg`.
+    2. Reference the `product-name` tag in your componentMap and use the new `data` prop available in the mapped component.
+```diff
+const defaultComponentsMap = {
++  product-name: (props) => <span>{props.data.product.name}</span>
+}
+```
+    <blockquote className="important">
+    <b>Note:</b> Please keep in mind that for simplicity sake, the product name component is not code split, but if it grows in complexity, please make sure to code split it using <code>@loadable/component</code> to avoid too large bundles.
```
real34, line 244:
```suggestion
    <b>Note:</b> Please keep in mind that for simplicity's sake, the product name component is not code splitted, but if it grows in complexity, please make sure to code split it using <code>@loadable/component</code> to avoid too large bundles.
```

source/docs/advanced/theme/wysiwyg.md
```diff
@@ -4,116 +4,385 @@ title: Display WYSIWYG content
...
-Widget.propTypes = {
-  type: PropTypes.string.isRequired,
-  attributes: PropTypes.object.isRequired
-};
+// ... rest of the file is intact
+```
+
+If you restart your application, you will notice that the new widget component is displayed. Hurray!
+
+However, in most cases, you will need to fetch data to display all the needed information in your widget. For instance, if the widget is `{% raw %}{{ widget type="acme/product-preview" sku="VSK12" }}{% endraw %}` you will want to fetch the product associated with the given SKU. This is the same concept as for the `WysiwygNodeData` but one level deeper in the GraphQL schema.
```
real34, line 313: This section feels like explaining the previous part again.
We may add a rough schema or a bold sentence highlighting the difference, otherwise I feel that readers might overlook this section.

source/docs/advanced/theme/wysiwyg.md
```diff
@@ -4,116 +4,385 @@ title: Display WYSIWYG content
...
+const defaultComponentsMap = {
+  h2: (props) => <h2 {...props} className="wysiwyg-h2" />
 }
 ```
 
-See [EnhanceWysiwyg's reference](/docs/reference/wysiwyg.html) for more details.
+The props available represent the HTML attributes coming from your Wysiwyg content. The only difference compared to traditional HTML is that `class` is renamed to `className` to better support React patterns and the `props.children` is an already constructed React element.
+
+<blockquote class="important">
+<b>Note:</b> If the component is complex, please think about code splitting them to avoid too large bundle on pages that don't need them.
```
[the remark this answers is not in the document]
JulienPradet: Since code splitting should only be using @loadable/component in the context of FC, I'd say we should wait for the accurate page documentation instead of pointing to a page that would point the readers in the wrong direction.

source/docs/advanced/theme/wysiwyg.md
JulienPradet: Not sure about splitted. https://www.usingenglish.com/reference/irregular-verbs/split.html

source/docs/advanced/theme/wysiwyg.md
JulienPradet: I suggest that we look into it later.

source/docs/advanced/theme/wysiwyg.md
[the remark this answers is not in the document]
real34: LGTM, and I noticed `@loadable/component` was mentioned later too.

PR #131, opened 2020-07-09
Add documentation about admin role detection in the front end

JulienPradet:


source/docs/advanced/graphql/dataloaders-and-cache-invalidation.md
```diff
@@ -406,6 +406,37 @@ export default {
 };
 ```
 
+### `PerMagentoAdminRole`
+
+_Since version 2.1.0_
+</blockquote>
+
+The `PerMagentoAdminRole` implementation is a decorator that is specific to Magento 2 module. It will decorate the existing caching strategies so that DataLoader keys are different for admin users and customers. This is useful if a custom Magento API returns a different result for admins and normal customers.
```
real34, line 415:
```suggestion
The `PerMagentoAdminRole` implementation is a decorator that is specific to Magento 2 integrations. It will decorate the existing caching strategies so that DataLoader keys are different for admin users and customers. This is useful if a custom Magento API returns a different result for admins and normal customers.
```

I wonder if it wouldn't make sense to start adding a tag system to features depending on the integrations where it is supported. Something like the `feature--new` with a visual tag-like system.

source/docs/magento2/detect-admin-users.md
```diff
@@ -0,0 +1,178 @@
+---
+id: detect-admin-users
+title: Detect admin users
+---
+
+_This feature has been added in version `2.1.0`_
+
+In some rare cases you could want to display things differently when the e-merchant is navigating your Front-Commerce application. For instance some pages may be protected and only accessible to the e-merchants of your shop. The rest of the users won't be able to read its content.
```
real34, line 8:
```suggestion
In some specific use cases you may want to adapt the application when the e-merchant is navigating your Front-Commerce application. For instance some pages may be protected and only accessible to the e-merchants of your shop. The rest of the users won't be able to read its content.
```

source/docs/magento2/detect-admin-users.md
```diff
@@ -0,0 +1,178 @@
...
+
+In Front-Commerce, we've achieved this by injecting something in the user's session when they are navigating Magento's backend. This documentation page details how it works and how you can use it.
+
+## Configuring your environment
+
+First, you need to make sure that you've configured the tokens that will enable the communication between Magento and Front-Commerce:
+
+- In Magento's store configuration, in General > General > Front-Commerce, fill in the "Magento Admin Token" field. It should be a random string of 32+ characters.
```
real34, line 16:
```suggestion
- In Magento's store configuration, under the "General > General > Front-Commerce" section, fill in the "Magento Admin Token" field. It should be a random string of 32+ characters.
```

source/docs/magento2/detect-admin-users.md
```diff
@@ -0,0 +1,178 @@
...
+
+    return {
+-      AdminData: AdminDataLoader(axiosInstance),
++      AdminData: AdminDataLoader(axiosInstance, loaders.MagentoAdmin),
+    };
+  },
+};
+```
+
+We can then use the `MagentoAdmin` loader to only return data when the admin is connected and use this information to authenticate the request. Please note that in this example, I've added a header using a specific token (`x-custom-header`). But this is an imaginary header for now and depends on how you have implemented `/admin-custom-endpoint` in your Magento. However the goal is to always have a protection on your Magento endpoint to make sure that only a request coming from Front-Commerce and with the correct information will be able to use this endpoint. Otherwise, if your Magento URL is compromised, any user could request sensitive admin data.
```
real34, line 111:
```suggestion
We can then use the `MagentoAdmin` loader to only return data when the admin is connected and use this information to authenticate the request. Please note that this example illustrates how to add a header using a specific token (`x-custom-header`). But this is an imaginary header for now and depends on how you have implemented `/admin-custom-endpoint` in your Magento. However the goal is to always have a protection on your Magento endpoint to make sure that only a request coming from Front-Commerce and with the correct information will be able to use this endpoint. Otherwise, if your Magento URL is compromised, any user could request sensitive admin data.
```

source/docs/magento2/detect-admin-users.md
```diff
@@ -0,0 +1,178 @@
...
++       return null;
++     }
+    },
+  };
+};
+```
+
+<blockquote class="important">
+**Important:** In case you rely on [Front-Commerce's caching mechanism](/docs/advanced/graphql/dataloaders-and-cache-invalidation.html) you will need to make sure that data fetched by admins are not cached in the same namespace as your normal users.
+For this reason, please make sure to update your [Caching strategies](/docs/advanced/graphql/dataloaders-and-cache-invalidation.html#Caching-strategies) with the relevant configurations.
```
real34, line 142:
```suggestion
For this reason, please make sure to update your [Caching strategies](/docs/advanced/graphql/dataloaders-and-cache-invalidation.html#PerMagentoAdminRole) with the relevant `PerMagentoAdminRole` configuration.
```

source/docs/magento2/detect-admin-users.md
```diff
@@ -0,0 +1,178 @@
...
+        return [
+            'adminId' => '13'
+        ];
+    }
+}
+```
+* If you are not in `developer` mode in magento, please make sure to rerun `bin/magento setup:di:compile` command.
+* You will then be able to use the following in Front-Commerce's server code for any authenticated admin users.
+```js
+console.log(loaders.MagentoAdmin.getAdminData().adminId)
```
real34, line 177:
```suggestion
const adminData = await loaders.MagentoAdmin.getAdminData();
console.log(adminData.adminId)
```

(related to https://gitlab.com/front-commerce/front-commerce/-/merge_requests/222#note_376339984)

PR #135, opened 2020-07-22
Document Affirm for Magento2 as a payment method

real34:
I also reordered payment methods so they are in an alphabetical order


real34:
> Looks good.
Thanks!

> Should we document your comment in the module's PR? What you've stated here seems different than the note about enabling the module.
> 
> > We'd have to be careful about developers that run module:enable all because it will fail if the Astound_Affirm module isn't installed. At some point we'll have to split everything in separate composer dependencies instead of a meta-package I think.

Not here, we may have to update the readme in the skeleton but it will very likely be migrated to this documentation at some point.
I'll add a note there while doing the 2.1.0 release.

PR #155, opened 2021-03-08
Update default Redis TTL documented

real34:


JulienPradet:
Shouldn't this change be documented in the migration guide? I believe it can have quite an impact on a shop if they didn't already have a reindex every night or something like that.

PR #157, opened 2021-03-11
Document how to setup a map

JulienPradet:
Since we now have multiple implementations of maps, it deserves its own page.


_config.yml
```diff
@@ -90,6 +90,7 @@ sidebar_categories:
   Shipping:
     - docs/advanced/shipping/overview
     - docs/advanced/shipping/custom-shipping-information
+    - docs/advanced/shipping/display-a-map
```
real34, line 93: I'd suggest that we move it under the more generic `Features` section, because developers may need it for a store locator or another use case not related to Shipping.

real34 (approved):
Thanks! I only have a suggestion about the page location in the menu :)

_config.yml
real34: I've done it in b740901

PR #170, opened 2021-04-14
fix: `front-commerce:performance` is documented twice

dpobel:


real34 (approved):
Well spotted! Thanks.

PR #173, opened 2021-04-26
2.6: Node minimum version

dpobel:
Migration guide + doc about Node.js minimum version

### DIFF
--- a/source/docs/appendices/migration-guides.md
+++ b/source/docs/appendices/migration-guides.md
@@ -9,6 +9,12 @@ Our goal is to make migrations as smooth as possible. This is why we try to make
 
 ## `2.5.0` -> `2.6.0`
 
+### Minimum Node.js version
+
+Node.js 10.x [reaches its end of life in the end of April 2021](https://nodejs.org/en/about/releases/). As a result, the required minimum Node.js version has been updated to the version 12.22.1.
+
+We also recommend to use Node.js 14.x as this version is also supported and it is the current active Long Term Support Node.js release.
+
 ### New icon required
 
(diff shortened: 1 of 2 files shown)


source/docs/appendices/migration-guides.md
```diff
@@ -9,6 +9,11 @@ Our goal is to make migrations as smooth as possible. This is why we try to make
 
 ## `2.5.0` -> `2.6.0`
 
+### Minimum Node.js version
+
+Node.js 10.x reaches the end of support in the end of April 2021. As a result,
```
real34, line 14: I suggest adding a direct link to the official documentation:
```suggestion
Node.js 10.x [reaches its end of life in the end of April 2021](https://nodejs.org/en/about/releases/). As a result,
```

source/docs/appendices/migration-guides.md
```diff
@@ -9,6 +9,11 @@ Our goal is to make migrations as smooth as possible. This is why we try to make
...
+the minimum Node.js required version has been updated to the version 12.22.1.
```
real34, line 15: It could be good to add a note that we recommend users to update their servers to 14.x directly (since this LTS is currently supported by FC). That way, when developers will check their node versions they could upgrade to an actively maintained node version!

real34 (approved):

source/docs/appendices/migration-guides.md
dpobel: done in 907eaeb

source/docs/appendices/migration-guides.md
dpobel: done in 907eaeb

PR #174, opened 2021-04-26
2.6: dependencies update notes

dpobel:
notes about the dependencies update

### DIFF
--- a/source/docs/appendices/migration-guides.md
+++ b/source/docs/appendices/migration-guides.md
@@ -9,6 +9,32 @@ Our goal is to make migrations as smooth as possible. This is why we try to make
 
 ## `2.5.0` -> `2.6.0`
 
+### Dependencies update
+
+In this release we have updated several dependencies. For most of them, the
+upgrade will be transparent. However, the following dependency updates require
+some attention though:
+
+* [`base-64`](https://www.npmjs.com/package/base-64) has been removed. It was
+    not used at all by Front-Commerce. If you use it, you have to make sure it's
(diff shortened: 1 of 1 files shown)


source/docs/appendices/migration-guides.md
```diff
@@ -9,6 +9,30 @@ Our goal is to make migrations as smooth as possible. This is why we try to make
...
+    might also warn you about new errors.
+* [`axios`](https://www.npmjs.com/package/axios) has been updated to 0.21.1.
+    Among other changes, it contains a fix that makes sure [the `@` character is
+    correctly URL encoded in URLs](https://github.com/axios/axios/issues/1212).
+    As a result, remote APIs now receive `%40` instead of a plain `@` when this
+    character is used in a query string parameter.
+* [`react-intl`](https://www.npmjs.com/package/react-intl) has been updated to
+    5.15.8. [The 5.x release has a minor backward incompatible](https://formatjs.io/docs/react-intl/upgrade-guide-5x)
+    compared to 4.x.
+
```
real34, line 35: It could be good to add a note that rel next/prev are now added by default by `react-paginate` (see https://gitlab.com/front-commerce/front-commerce/-/merge_requests/463#note_555995920)

real34 (approved):
I've only added a suggestion. Otherwise it looks good. Thanks for keeping track of all of these changes.

Note: we found that when writing documentation it was much easier to collaborate on markdown that didn't had returns as it was easier to work on suggestions and apply them. That's the reason why most of the content doesn't have auto returns after a given length.

PR #179, opened 2021-04-30
Clarify version of dependencies for 2.6

real34:
This PR rewords a sentence about the Node version bump (from user feedback) and adds the related "Release notes" section for 2.6.

BTW, Nginx supports Stale-While-Revalidate [since NGINX 1.11.10](https://www.nginx.com/blog/nginx-caching-guide/) so I've also updated the minimum Nginx documented version.
I used a ~2yo version that matches the Nginx plus LTS.


dpobel (approved):
:+1:

PR #180, opened 2021-05-04
doc: Magento1 setup requires Add Store Code to Urls to be disabled

dpobel:
this problem behind the recent support issue…


real34 (approved):
Thanks! Well spotted!

PR #183, opened 2021-05-07
doc: migration guide for new Elasticsearch facets behavior

dpobel:


source/docs/appendices/migration-guides.md
```diff
@@ -9,6 +9,14 @@ Our goal is to make migrations as smooth as possible. This is why we try to make
 
 ## `2.6.0` -> `2.7.0`
 
+### More reliable facets generation with Elasticsearch
+
+As off the 2.7 version, Front-Commerce fetches the attributes on which facets can be generated from Magento (for both version 1 and 2). Previously, the implementation was relying on an attribute added in the Elasticsearch index by the ElasticSuite module to figure out which ones can be used to generate facets. This could lead to wrong Elasticsearch queries and it was common to have to ignore some attributes using [the `search.ignoredAttributeKeys` configuration](/docs/reference/configurations.html#config-website-js). As a result, when upgrading to 2.7:
```
real34, line 14:
```suggestion
As of the 2.7 version, Front-Commerce fetches the attributes on which facets can be generated from Magento (for both version 1 and 2). Previously, the implementation was relying on an attribute added in the Elasticsearch index by the ElasticSuite module to figure out which ones can be used to generate facets. This could lead to wrong Elasticsearch queries and it was common to have to ignore some attributes using [the `search.ignoredAttributeKeys` configuration](/docs/reference/configurations.html#config-website-js). As a result, when upgrading to 2.7:
```

source/docs/appendices/migration-guides.md
```diff
@@ -9,6 +9,14 @@ Our goal is to make migrations as smooth as possible. This is why we try to make
...
+
+1. make sure the attributes on which you want some facets are configured so that _Use in Layered Navigation_ is set to _Filterable (with results)_
+1. if you need to have an attribute configured this way but still want to ignore it in the facets, you can add it to the `search.ignoredAttributeKeys` configuration
+1. you can safely remove attribute codes from `search.ignoredAttributeKeys` that were added to prevent wrong Elasticsearch queries (`description`, `short_description`,…)
```
real34, line 18: This should be double-checked for all of their custom attributes, it is not really something to do blindly by thinking it is "safe".

```suggestion
1. you can remove attribute codes that are not "Used in Layered Navigation" from the `search.ignoredAttributeKeys` configuration. They were probably text fields added here to prevent incorrect Elasticsearch queries (`description`, `short_description`,…)
```

real34 (approved):
Thanks!

PR #185, opened 2021-05-07
 Document the new configurations to send additional headers in Magento API calls

real34:
We don't have any reference of all available configurations from the providers.

Hence I'm not really sure of where to add such things. If you have a better idea, I'm open to suggestions!


source/docs/appendices/migration-guides.md
```diff
@@ -9,6 +9,12 @@ Our goal is to make migrations as smooth as possible. This is why we try to make
 
 ## `2.6.0` -> `2.7.0`
 
+### Additional headers in Magento API calls
+
+Both Magento 1 and Magento 2 modules in Front-Commerce now supports sending additional headers in all API calls. To do so, you must define the `magento.api.extraHeaders` (for storefront API) and/or `magento.api.extraAdminHeaders` (for admin API) configuration values from a [configuration provider](/docs/advanced/server/configurations.html#What-is-a-configuration-provider).
```
dpobel, line 14: IMHO, that does not belong to the migration guide as there's nothing to migrate
on that topic. I would rather put that under Magento1 and Magento2 section maybe
in "Advanced" page ?

real34: I agree. We could start with an "Advanced" page with miscellaneous and split it later if needed.

I created them instead of the migration guide:
- https://deploy-preview-185--heuristic-almeida-1a1f35.netlify.app/docs/magento1/advanced.html
- https://deploy-preview-185--heuristic-almeida-1a1f35.netlify.app/docs/magento2/advanced.html

dpobel (approved):

PR #199, opened 2021-06-10
useEmailServerValidation documentation

dpobel:
and fixes some typos and typography issues

### DIFF
--- a/source/docs/advanced/features/smart-forms.md
+++ b/source/docs/advanced/features/smart-forms.md
@@ -254,7 +254,7 @@ import { Email } from "theme/components/atoms/Form/Input";
      </Form>
 ```
 
-With this code, when the user submits the form, the hook will call the GraphQL API to validate the email. The GraphQL email validation API is designed [to return different validation statuses](https://gitlab.com/front-commerce/front-commerce/-/merge_requests/546/diffs#e987e1da0e06e52d1b120c9d0081953f11fc8c7a_13_18) (not only valid/invalid) and by default, the `onSubmit` handler returned by `useEmailServerValidation` has the following behaviour:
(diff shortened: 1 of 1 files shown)


source/docs/advanced/features/smart-forms.md
```diff
@@ -217,3 +217,102 @@ const MyAwesomeForm = ({
...
+
+   return (
+-    <Form onValidSubmit={submit}>
++    <Form onValidSubmit={onSubmit}>
+        <Email name="email" id="email" value={email} required />
+        <!-- probably some other fields -->
+     </Form>
+```
+
+With this code, when the user submits the form, the hook will call the GraphQL API to validate the email. The GraphQL email validation API is designed [to return different validation statuses](https://gitlab.com/front-commerce/front-commerce/-/merge_requests/546/diffs#e987e1da0e06e52d1b120c9d0081953f11fc8c7a_13_18) (not only valid/invalid) and by default, the `onSubmit` handler returned by `useEmailServerValidation` has the following behaviour:
```
real34, line 257: Nit: the MR will remain as is while the code will stay up-to-date. While it can also be refactored, in this context it seems more reliable than the MR
```suggestion
With this code, when the user submits the form, the hook will call the GraphQL API to validate the email. The GraphQL email validation API is designed [to return different validation statuses](https://gitlab.com/front-commerce/front-commerce/-/blob/main/src/server/modules/front-commerce/smart-forms/loaders/EmailValidation.js#L4) (not only valid/invalid) and by default, the `onSubmit` handler returned by `useEmailServerValidation` has the following behaviour:
```

real34 (approved):
:+1: Thanks for these changes.

source/docs/advanced/features/smart-forms.md
dpobel: that makes sense but then maybe the link should point to the GraphQL schema which is probably the less likely to change

PR #205, opened 2021-07-16
Document how to disable server response compression

real34:
Feature introduced in https://gitlab.com/front-commerce/front-commerce/-/merge_requests/615

### DIFF
--- a/source/docs/appendices/migration-guides.md
+++ b/source/docs/appendices/migration-guides.md
@@ -7,6 +7,16 @@ This area will contain the Migration steps to follow for upgrading your store to
 
 Our goal is to make migrations as smooth as possible. This is why we try to make many changes backward compatible by using deprecation warnings. The deprecation warnings are usually removed in the next breaking release.
 
+## `2.7.0` -> `2.8.0`
+
+### Disable response compression
+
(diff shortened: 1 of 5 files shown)


source/docs/appendices/migration-guides.md
```diff
@@ -7,6 +7,16 @@ This area will contain the Migration steps to follow for upgrading your store to
 
 Our goal is to make migrations as smooth as possible. This is why we try to make many changes backward compatible by using deprecation warnings. The deprecation warnings are usually removed in the next breaking release.
 
+## `2.7.0` -> `2.8.0`
+
+### Disable response compression
```
dpobel, line 12: remark (not really specific to this MR, we've done that several times I think): we kind of use the migration guide to "announce" new features and here the feature documentation ends up being (almost) only in the migration guide.

While this is probably a good thing for people upgrading, the documentation of this feature will be hard to find/discover for new project. I would suggest to move this paragraph in a page under https://developers.front-commerce.com/docs/advanced/performance/overview.html. This new page should be linked in the release announcement and maybe in the migration guide of each release we could also add a _New feature_ section so that way we would satisfy both the upgrade use case and the new project one.

real34: That is a very good point.

I gave it a first try in 24f972b.

dpobel (approved):

PR #208, opened 2021-07-20
added prismic routable&sitemapable docs

mo-fc:


source/docs/prismic/routable-types.md
```diff
@@ -0,0 +1,220 @@
...
+id: routable-types
+title: Prismic Routable Types
+---
+
+To make a custom Prismic data type routable these steps are required:
+
+- [Create a field to represent the Url of the custom type](#create-a-field-to-represent-the-url-of-the-custom-type)
+- [Create a GraphQL type to represent the custom type](#create-a-graph-ql-type-to-represent-the-custom-type)
+- [Register the Prismic custom type as a routable type](#register-the-prismic-custom-type-as-a-routable-type)
+- [Map GraphQL type to a component](#map-graph-ql-type-to-a-component)
```
dpobel, line 11: as I was saying on Slack, the anchors do not work (mostly case issues) it seems

See https://deploy-preview-208--heuristic-almeida-1a1f35.netlify.app/docs/prismic/routable-types.html

source/docs/prismic/routable-types.md
```diff
@@ -0,0 +1,220 @@
...
++        }
++      }
++    });
+
+    return {};
+  },
+};
+```
+
+Note the `postTransformer` above is optional. It is a function that will be called after the transformation is done using `contentTransformOptions`. it is given the current URL being resolved and the transformed document. It can be used if you have some custom logic to apply to the transformed document or if you want to supress a document from showing using some custom logic (returning a falsey value).
```
dpobel, line 103:
```suggestion
Note the `postTransformer` above is optional. It is a function that will be called after the transformation is done using `contentTransformOptions`. It is given the current URL being resolved and the transformed document. It can be used if you have some custom logic to apply to the document or if you want to suppress a document from showing using some custom logic (returning a _falsy_ value).
```

source/docs/prismic/routable-types.md
```diff
@@ -0,0 +1,220 @@
...
+    },
++    priority: () => 1,
++    seoImages: (content) => [],
++    lastmod: () => "2020-01-01T12:00:00.000Z",
++    changefreq: () => "daily",
+  },
+};
+```
+
+2. Enable sitemap in the module definition file:
```
dpobel, line 181: wrong number in the resulting page https://deploy-preview-208--heuristic-almeida-1a1f35.netlify.app/docs/prismic/routable-types.html

I think you need to get rid of the empty lines and add an indeed level to code blocks

source/docs/prismic/routable-types.md
```diff
@@ -0,0 +1,220 @@
...
++    },
++  },
++};
+```
+
+Note we added a path field resolver that appends the shop's baseUrl  to the `url` field of our custom Prismic type.
+
+## Register the Prismic custom type as a routable type
+
+To make a custom Prismic type routable it must be registered with the Prismic module. First ensure that the `Prismic/core` is added as a module's dependency then use `loaders.Prismic.registerRoutableType` to add the custom type as routable type:
```
dpobel, line 63:
```suggestion
To make a custom Prismic type routable it must be registered with the Prismic module. First ensure that the `Prismic/Core` is added as a module dependency then use `loaders.Prismic.registerRoutableType` to add the custom type as a routable type:
```

dpobel (approved):
:+1:

PR #211, opened 2021-07-21
 Document SSR pages restriction from 2.8

real34:
This MR documents the change brought in https://gitlab.com/front-commerce/front-commerce/-/merge_requests/623

### DIFF
--- a/source/docs/advanced/theme/server-side-rendering.md
+++ b/source/docs/advanced/theme/server-side-rendering.md
@@ -123,6 +123,20 @@ Hopefully the previous screen will prevent such errors to occur in production, b
 
 If you still want to display the `theme/pages/SsrFallback` page in dev mode, [add `FRONT_COMMERCE_DEV_SSR_FALLBACK_DISABLE=true` to your environment variables](/docs/reference/environment-variables.html#DX).
 
+## Restrictions on pages that will be server-rendered
+
+<blockquote class="feature--new">
+  _This feature has been added in version `2.8.0`_
+</blockquote>
+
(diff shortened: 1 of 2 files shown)


source/docs/advanced/theme/server-side-rendering.md
```diff
@@ -123,6 +123,20 @@ Hopefully the previous screen will prevent such errors to occur in production, b
...
+  _This feature has been added in version `2.8.0`_
+</blockquote>
+
+Front-Commerce avoids to unnecessary render a full 404 page server-side for URLs that don't need it. When bots access random URLs (e.g: `dump.sql.gz`, `password.txt` etc…) or outdated static assets, Front-Commerce will return a simpler 404 error.
+
+Pages are server-rendered when accessed with the following URL patterns:
+- without extension (e.g: `/contact`, `/faq/our-return-policy`…)
+- with an `.html` extension (e.g: `/contact.html`, `/wooden-table-with-chairs.html`…)
+
+**If your application contains pages whose URL have extensions other than `.html`, please contact us.**
```
dpobel, line 138: a bit late feedback, but we could also add `.htm` to the list

dpobel (approved):

source/docs/advanced/theme/server-side-rendering.md
real34: I prefer that we wait for use cases in projects.\
The current reason why `.html` is supported is because that is the default Magento setting and that projects have it for legacy reasons.

`.htm` would then result from either a new platform, a very specific use case or a custom Magento config… we'll have to decide if we make it configurable for devs, keep it hardcoded, fetch remote configuration or something else.

PR #215, opened 2021-07-22
Improve algolias documentation

dpobel:
from customer feedback

### DIFF
--- a/source/docs/magento1/search-engine.md
+++ b/source/docs/magento1/search-engine.md
@@ -176,3 +176,8 @@ query Search {
 
 If you are using the default theme or the theme Chocolatine, the search bar
 should now be visible.
+
+
+<blockquote class="warning">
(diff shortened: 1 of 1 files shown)


source/docs/magento1/search-engine.md
```diff
@@ -100,6 +100,8 @@ After the installation, you need to configure it, at least:
    section, set the _Enable Indexing_ option
 1. in the same section, fill the credentials you can find on [the Algolia
    Dashboard](https://www.algolia.com/dashboard/api-keys)
+1. The attribute `category_ids` must be listed in both the facets configuration and in the indexed product attributes (and there it must be set as _Searchable_)
```
real34, line 103: We should add a sentence explaining why this is needed. In context that don't use Algolia for Category pages it is not required.

```suggestion
1. The attribute `category_ids` must be listed in both the facets configuration and in the indexed product attributes (and there it must be set as _Searchable_) if you plan to use Algolia for category pages and facets
```

real34 (approved):

source/docs/magento1/search-engine.md
dpobel: fixed in 2cc04e3

PR #216, opened 2021-07-22
store urls should not have trailing slashes

mo-fc:


real34:
The content is too vague, and from my understanding I don't think it is correct.

FC doesn't accept store URLs with a trailing slash: https://gitlab.com/front-commerce/front-commerce/-/blob/2ce919c810f582f04e829ec95cbe4abd81afcf47/src/server/core/config/shopConfigProvider.js#L127

real34 (changes requested):
The content is too vague, and from my understanding I don't think it is correct.

FC doesn't accept store URLs with a trailing slash: https://gitlab.com/front-commerce/front-commerce/-/blob/2ce919c810f582f04e829ec95cbe4abd81afcf47/src/server/core/config/shopConfigProvider.js#L127

mo-fc:
I know it should have thrown an error no? or is it an issue with the platform.sh router?

PR #220, opened 2021-07-22
Document how to refresh the prismic ref without restarting FC

dpobel:
using cURL as it's a bit more common than httpie

### DIFF
--- a/source/docs/prismic/installation.md
+++ b/source/docs/prismic/installation.md
@@ -34,7 +34,7 @@ In case of trouble, `front-commerce:prismic` can be added to the `DEBUG` environ
 </blockquote>
 
 <blockquote class="tip">
-Configuring a webhook is required so that Front-Commerce fetches the last version of the Prismic content. In a development environment, you might skip this step. To make sure your running Front-Commerce uses the last version of the Prismic content without restarting it, you can use the following _oneliner_ from the skeleton directory to simulate a webhook call:
+Configuring a webhook is required so that Front-Commerce fetches the last version of the Prismic content. In a development environment, you might skip this step. To make sure your currently running Front-Commerce instance uses the lastest version of the Prismic content without restarting it, you can use the following _oneliner_ from the skeleton directory to simulate a webhook call:
 
 ```sh
 source .env ; PAYLOAD=`printf '{"secret": "%s", "type": "api-update", "masterRef": true}' $FRONT_COMMERCE_PRISMIC_WEBHOOK_SECRET` ; curl -H "Content-Type: application/json" -d "$PAYLOAD" $FRONT_COMMERCE_URL/prismic/webhook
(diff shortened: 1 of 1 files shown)


source/docs/prismic/installation.md
```diff
@@ -33,8 +33,12 @@ FRONT_COMMERCE_PRISMIC_WEBHOOK_SECRET=a-secret-defined-in-webhook-configuration
 In case of trouble, `front-commerce:prismic` can be added to the `DEBUG` environment variable value, this value turns the debug on for Prismic module and make it verbose.
 </blockquote>
 
-<blockquote class="note">
-Configuring a webhook is required so that Front-Commerce fetches the last version of the Prismic content. In a development environment, you might skip this step but in that case, you will have to restart Front-Commerce after each publish operation in Prismic.
+<blockquote class="tip">
+Configuring a webhook is required so that Front-Commerce fetches the last version of the Prismic content. In a development environment, you might skip this step. To make sure your currently running Front-Commerce instance uses the lastest version of the Prismic content without restarting it, you can use the following _oneliner_ from the skeleton directory to simulate a webhook call:
+
+```sh
+source .env ; PAYLOAD=`printf '{"secret": "%s", "type": "api-update", "masterRef": true}' $FRONT_COMMERCE_PRISMIC_WEBHOOK_SECRET` ; curl -H "Content-Type: application/json" -d "$PAYLOAD" $FRONT_COMMERCE_URL/prismic/webhook
```
real34, line 40: We can keep the one liner that might be convenient for users however I think we also must document a good ol' simple curl commands that could be understood by integrators at a first glance.

The reason for this is that we've often found that they use Postman (or similar tool) to save this kind of "common" requests. They usually transpose it directly from the doc to the UI.

The `<blockquote>` might getting too big, and maybe adding a sub-section would make it more readable while also allowing direct linking using its anchor (useful for support or doc).

real34 (changes requested):

source/docs/prismic/installation.md
mo-fc: curl command generated from postman:

```suggestion
source .env ; PAYLOAD=`printf '{"secret": "%s", "type": "api-update", "masterRef": true}' $FRONT_COMMERCE_PRISMIC_WEBHOOK_SECRET` ; curl -H "Content-Type: application/json" -d "$PAYLOAD" $FRONT_COMMERCE_URL/prismic/webhook

curl --location -g --request POST '{FRONT_COMMERCE_URL}/default/prismic/webhook' \
--header 'Content-Type: application/json' \
--data-raw '{
    "secret": FRONT_COMMERCE_PRISMIC_WEBHOOK_SECRET, "type": "api-update", "masterRef": true
}'
```

mo-fc: I committed the above suggestion

I think it should be good now

dpobel: I think Pierre meant something a bit more elaborated in a dedicated paragraph where we should explain the webhook only does a POST request with a specific body that can be done with any HTTP client and then give an example with curl with visible fake data and then the one liner that does everything.

dpobel: done in fb08079

real34 (approved):
That's indeed much better IMO!

Thanks for the rewrite.

Issue #223, opened 2021-07-27
Document routable pages and new Dispatcher


real34:
We've detected that the [Handle dynamic URLs with the Dispatcher
](https://developers.front-commerce.com/docs/advanced/theme/route-dispatcher.html) is outdated.

It must be rewritten/improved to explain how to achieve this with the "new" routing system.\
See https://real34.github.io/slides/routing-with-graphql-interfaces/#40 our the training slides as source (ask the team for a link).

PR #228, opened 2021-08-02
update dispatcher url matcher docs

mo-fc:
the documentation for the dispatcher was out of date and not documenting the registerUrlMarcher from loaders.Page

### DIFF
--- a/source/docs/advanced/theme/route-dispatcher.md
+++ b/source/docs/advanced/theme/route-dispatcher.md
@@ -58,66 +58,121 @@ const moduleDefinition = {
 export default moduleDefinition;
 ```
 
-```diff
+```js
 // urlMatcher.js
-+const myCustomUrlMatcher = (loader) => async (url, match) => {
-+  if(match.__typename) {
-+    // a match have been already loaded (you can still override it or return it as is like we do here)
-+    return match;
-+  }
-+  const myMatch = await loader.loadByUrl(url);
-+  if(!myMatch) {
-+    return match; // return original match if none is found (this is needed!)
-+  }
-+  return { // return your custom routable entity
-+    path: <canonical_url>,      // depending on your use case can be same as url
-+    url,
-+    __typename: 'MyCustomType', // the GraphQL type of your custom type (should be added in schema.gql),
-+    ...myMatch, // or some adaptation of it. this will be available to your component defined bellow on a prop called `matched`
-+  }
-+};
-+
-+export default myCustomUrlMatcher;
+const myCustomUrlMatcher = (loader) => async (url, match) => {
+  if (match.__typename) {
+    // a match have been already loaded (you can still override it or return it as is like we do here)
+    return match;
+  }
+  const myMatch = await loader.loadByUrl(url);
+  if (!myMatch) {
+    return match; // return original match if none is found (this is needed!)
+  }
+  return {
+    // return your custom routable entity
+    path: myMatch.canonical_url || url, // depending on your use case can be same as url
+    url,
+    __typename: "MyCustomType", // the GraphQL type of your custom type (should be added in schema.gql),
+    ...myMatch, // or some adaptation of it. this will be available to your component defined bellow on a prop called `matched`
+  };
+};
+
+export default myCustomUrlMatcher;
 ```
 
-```diff
+```gql
 # schema.gql
-+MyCustomType {
-+  title: String! # example field
-+  ... # define own schema
-+}
+MyCustomType implements Routable {
+  path: String! # needed for Routable
+  title: String! # example field
+  ... # define own schema
+}
 ```
 
-```diff
+```js
 // loader.js
 const MyCustomLoader = (axiosInstance) => {
   return {
-+    loadByUrl: (url) => {
-+      return ( // example: get data by issuing an external api call
-+        axiosInstance.get(...)
-+        .then((response) => response.data)
-+        .catch((err) => err.response?.status === 404 ? null : Promise.reject(err));
-+      );
-+    },
+    loadByUrl: (url) => {
+      return (
+        // example: get data by issuing an external api call
+        axiosInstance
+          .get(`/my/api/my-custom-entity?url=${encodeURIComponent(url)}`)
+          .then((response) => response.data)
+          .catch((err) =>
+            err.response?.status === 404 ? null : Promise.reject(err)
+          )
+      );
+    },
   };
 };
 ```
 
-    <blockquote class="note">
+<blockquote class="note">
     See [Loaders](/docs/advanced/graphql/slim-down-resolvers-with-loaders.html) documentation to learn how to instantiate your loader and add it to your GraphQL context.
-    </blockquote>
+</blockquote>
 
-Once you've done, you should be able to test that everything works as expected by using the GraphQL Playground at http://localhost:4000/playground and executing the following query:
+Once this is done, you should be able to test that everything works as expected by using the GraphQL Playground at http://localhost:4000/playground and executing the following query:
 
 ```gql
 {
   route(url: "some-dynamic-custom-url") {
-    url
-    title # example
-    ...   # custom type fields
+    path
+    ... on MyCustomType {
+      title # example
+    }
   }
 }
 ```
+
+## Add GraphQL type to the dispatcher query
+
+Override `DispatcherQuery` if you have not already done so. You can find it under `src/web/theme/modules/Router/DispatcherQuery.gql`.
+
+Add an [inline fragment](https://graphql.org/learn/queries/#inline-fragments) to the `DispatcherQuery` for your newly created GraphQL type:
+
+```diff
+...
++import 'path_to_your_fragment/MyCustomTypeFragment.gql'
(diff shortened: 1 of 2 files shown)


real34:
Note: related to #223

source/docs/advanced/theme/route-dispatcher.md
```diff
@@ -22,103 +22,118 @@ If you come from a Magento background, this is the concept behind [URL Rewrites]
...
-            return result;
-          } else {
-            return loaders.Url.matchBy(url);
-          }
-        })
-    },
-  }
+```diff
+# schema.gql
++MyCustomType {
```
real34, line 85: This type must implements the `Routable` interface and thus have à `path: String` field.

I think it should be part of the example.

source/docs/advanced/theme/route-dispatcher.md
```diff
@@ -22,103 +22,118 @@ If you come from a Magento background, this is the concept behind [URL Rewrites]
...
-  }
-}
+export default moduleDefinition;
 ```
-    <blockquote class="note">
-    See [Loaders](/docs/advanced/graphql/slim-down-resolvers-with-loaders.html) documentation to learn how to instantiate your loader and add it to your GraphQL context.
-    </blockquote>
 
-    Here, we only handle `my-dynamic-custom-url` URL. However, you will most likely need to fetch the result from one of your backend instead. If you do so, the final result should still match the structure of an Entity as presented above.
+```diff
```
real34, line 61: For whole files, I can't see any benefits to `diff` syntax.
Here, `js` might lead to more readable example IMO. I suggest to rework the example.

source/docs/advanced/theme/route-dispatcher.md
```diff
@@ -22,103 +22,118 @@ If you come from a Magento background, this is the concept behind [URL Rewrites]
...
+// urlMatcher.js
++const myCustomUrlMatcher = (loader) => async (url, match) => {
++  if(match.__typename) {
```
real34, line 64: CS in this example needs to be reworked.
We must try to keep it as standard as possible.

I suggest at least:
- pass it into Prettier (either in your editor or in https://prettier.io/playground/)
- ensure the code is _valid_ (e.g: `<canonical url>` should instead use a property of `myMatch`)
- prevent looooooong lines: either remove comments (over a paragraph below) or put them on another line(s)… refactoring into temporary variables can also make things more expressive and remove the need for some comments

source/docs/advanced/theme/route-dispatcher.md
```diff
@@ -22,103 +22,118 @@ If you come from a Magento background, this is the concept behind [URL Rewrites]
...
++  ... # define own schema
++}
+```
+
+```diff
+// loader.js
+const MyCustomLoader = (axiosInstance) => {
+  return {
++    loadByUrl: (url) => {
++      return ( // example: get data by issuing an external api call
```
real34, line 96: Same feedbacks as above about CS in this code (indentation and _executable_ call). It could also be a `js` block instead of a `diff`.

Here for instance, the comment could be removed if an expressive URL is used in `axiosInstance.get` instead of `...` (suggestion: `/my/api/my-custom-entity/${url}`).

source/docs/advanced/theme/route-dispatcher.md
```diff
@@ -22,103 +22,118 @@ If you come from a Magento background, this is the concept behind [URL Rewrites]
...
++    },
+  };
 };
 ```
-    Note that we don't want to forget to call the initial `loaders.Url.matchBy` here. If we don't, we will break the base functionalities of Front-Commerce.
 
-    Additionally, since we are using an existing loader, we must make sure that our dependencies are correctly configured in our module. In case of a store using Magento, we should add the dependency on `Magento2/Url` module. See [Graphql module definition Reference](/docs/reference/graphql-module-definition.html#dependencies-optional) for more details.
+    <blockquote class="note">
+    See [Loaders](/docs/advanced/graphql/slim-down-resolvers-with-loaders.html) documentation to learn how to instantiate your loader and add it to your GraphQL context.
+    </blockquote>
```
real34, line 108:
```suggestion
<blockquote class="note">
    See [Loaders](/docs/advanced/graphql/slim-down-resolvers-with-loaders.html) documentation to learn how to instantiate your loader and add it to your GraphQL context.
</blockquote>
```

Indentation led to an automatic code block instead:
Selection_001

source/docs/advanced/theme/route-dispatcher.md
```diff
@@ -22,103 +22,118 @@ If you come from a Magento background, this is the concept behind [URL Rewrites]
...
 
-Once you've done these three steps, you should be able to test that everything works as expected by using the GraphQL Playground at http://localhost:4000/playground and executing the following query:
+Once you've done, you should be able to test that everything works as expected by using the GraphQL Playground at http://localhost:4000/playground and executing the following query:
```
real34, line 110:
```suggestion
Once this is done, you should be able to test that everything works as expected by using the GraphQL Playground at http://localhost:4000/playground and executing the following query:
```

source/docs/advanced/theme/route-dispatcher.md
```diff
@@ -22,103 +22,118 @@ If you come from a Magento background, this is the concept behind [URL Rewrites]
...
 {
-  route(url: "my-dynamic-custom-url") {
+  route(url: "some-dynamic-custom-url") {
     url
-    type
-    identifier
-    target_path
-    redirect_type
+    title # example
+    ...   # custom type fields
```
real34, line 117:
```suggestion
```
not needed as the `title` should be sufficient. It makes the example executable without syntax error.

source/docs/advanced/theme/route-dispatcher.md
```diff
@@ -22,103 +22,118 @@ If you come from a Magento background, this is the concept behind [URL Rewrites]
...
 ```gql
 {
-  route(url: "my-dynamic-custom-url") {
+  route(url: "some-dynamic-custom-url") {
     url
-    type
-    identifier
-    target_path
-    redirect_type
+    title # example
```
real34, line 116: `path` is the only field from the `Routable` interface.

This example should thus be reworked with something close to:

```graphql
{
  route(url: "some-dynamic-custom-url") {
    path
	... on MyCustomType {
	    title # from the example schema above
 	}
  }
}
```

real34 (changes requested):
Several changes are needed to make it correct and consistent with the rest of the documentation.

There is also a section missing: updating the Dispatcher query.
We may reuse content from https://developers.front-commerce.com/docs/prismic/routable-types.html#Add-GraphQL-type-to-the-dispatcher-query

BTW, the example in Prismic could be simplified and a link added to this page if readers want more details.

source/docs/advanced/theme/route-dispatcher.md
```diff
@@ -22,103 +22,173 @@ If you come from a Magento background, this is the concept behind [URL Rewrites]
...
+```diff
+import typeDefs from "./schema.gql";
++import myCustomUrlMatcher from './urlMatcher.js';
+import MyCustomLoader from './loader.js';
+import { makeUserClientFromRequest } from "server/modules/magento2/core/factories";
+
+const moduleDefinition = {
+  namespace: "MyCustomModule",
+-  dependencies: ["Front-Commerce/Core"],
++  dependencies: [],
```
real34, line 46: It should be the opposite:
```suggestion
-  dependencies: [],
+  dependencies: ["Front-Commerce/Core"],
```

source/docs/advanced/theme/route-dispatcher.md
```diff
@@ -22,103 +22,173 @@ If you come from a Magento background, this is the concept behind [URL Rewrites]
 In Front-Commerce’s core integrations (such as Magento2), the association between a URL and a page is already implemented for entities like Products, Categories, CMS pages… But depending on your own site, you might need to add new ones.
 
 To do so, you will need to proceed in two steps:
-* Support the new URLs in the `route` query in your GraphQL Schema
-* Add the mapping between the type returned by `route` and the page component that should be displayed
 
-## Support the new URLs in the `route`
+- Register a new urlMatcher
+- Add the mapping between the type returned by `route` and the page component that should be displayed
```
real34, line 27:
```suggestion
- Add GraphQL type to the dispatcher query
- Add the mapping between the type returned by `route` and the page component that should be displayed
```

It'd be perfect with anchor links to the related section too.

source/docs/advanced/theme/route-dispatcher.md
```diff
@@ -22,103 +22,173 @@ If you come from a Magento background, this is the concept behind [URL Rewrites]
...
++      ...MyCustomTypeCacheControlFragment # optional check note below for link
++    }
   }
 }
 ```
 
-## Add the mapping between the `type` and the page component
+```gql
+# MyCustomTypeFragment.gql
+fragment AlbumFragment on Album {
```
real34, line 169:
```suggestion
fragment MyCustomTypeFragment on MyCustomType {
```

real34 (approved):
Thanks! That's much better.

I detected some minor typos but :+1:

PR #235, opened 2021-08-23
fix: wrong method name and typos in analytics

dpobel:
I noticed the wrong method name (`page` instead of `event`) in the example while answering a support question. It also fixes some typos

### DIFF
--- a/source/docs/advanced/theme/analytics.md
+++ b/source/docs/advanced/theme/analytics.md
@@ -15,7 +15,7 @@ If we represent how it works, it would look like this:
 
 Across your React application, you can track events using functions such as `trackEvent` or `trackPage`. Then, the event is dispatched to all the relevant integrations registered in your application.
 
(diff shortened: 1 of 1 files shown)


source/docs/advanced/theme/analytics.md
```diff
@@ -199,7 +199,7 @@ However, please note that some integrations are not up to date. Even if you find
 
 ### Main integrations
 
-Here is a list of integrations frequently used accross eCommerce shops:
+Here is a list of integrations frequently used across e-commerce shops:
```
real34, line 202: This one is touchy :)

We got the discussions several times over the year and decided to not care.
I'm guilty of all the eCommerce as we first decided this (and took the habit).

dpobel: ahah 
and for the whole story, the spellchecker did not like it :)

PR #239, opened 2021-09-30
added configurable options migration guides

mo-fc:


source/docs/appendices/migration-guides.md
```diff
@@ -7,6 +7,58 @@ This area will contain the Migration steps to follow for upgrading your store to
...
+- The product
+- The initial selected options (with one of the following formats `{ [optionId]: valueId }` or `{ [attribute or label], [value or optionValue] }`)
+
+And returns an Object with the following properties:
+
+- selectedProduct: The selected product (does not change the sku of the input product)
+- setOption: A function to set an option
+- selectedOptions: The current selected options in the format `{ [optionId]: valueId }`
+- baseProduct: The base product (that the selected product is part of)
+- selectedSku: The sku if the selected product
```
real34, line 25: This is a migration guide.
It should remain focused on the migration process.

I suggest to only keep an introduction sentence to provide context:
```suggestion
In this release we have done a refactor of the configurable product options. We have extracted much of the logic from configurable product options from HOCs into functions that can be used even outside of react.

This refactoring homogeneized how options were used and allowed us to fix some discrepancies betweent the product page and cart item update configurators. Internal APIs were updated to cope with this change, and we deprecated some legacy props.
```

The rest could go in the _REFERENCE_ section to document the API. A link could be added to a [New features in `2.10.0`](https://developers.front-commerce.com/docs/appendices/migration-guides.html#New-features-in-2-9-0] section.

real34 (changes requested):
The migration details are good, thanks!

My only concern is about the introduction that is too detailed for a migration guide. I've suggested a shorter version and to move it elsewhere in  the doc.

source/docs/appendices/migration-guides.md
mo-fc: interesting pattern! I updated as mentioned above please have a look and approve if all is well

real34 (approved):
Thanks! It looks better :+1:

PR #241, opened 2021-10-06
wrong package name

mo-fc:
no `front-commerce/analytics.js-integration-google-tag-manager`


real34 (changes requested):
The documentation is correct.
We have forks for them:
- https://github.com/front-commerce/analytics.js-integration-google-tag-manager/commits/master
- https://github.com/front-commerce/analytics.js-integration-google-analytics/commits/master

PR #251, opened 2021-10-28
Migration guide for zoom in product image feature

dpobel:
https://deploy-preview-251--heuristic-almeida-1a1f35.netlify.app/docs/appendices/migration-guides.html#Zoom-in-on-product-images


real34 (approved):
Thanks for these changes. It should be good now.

PR #252, opened 2021-10-29
Add codemod cli documentation

paul-vd:


paul-vd:
Closed because based of off wrong branch

PR #254, opened 2021-11-02
feat(webpack5): update the storybook documentation

abailleulfc:


source/docs/appendices/migration-guides.md
```diff
@@ -20,6 +20,16 @@ These new features may be relevant for your existing application:
...
 - Front-Commerce dependencies are now regularly (and automatically) updated using [Depfu](https://depfu.com/). Patches are automatically applied if the CI is green, minor versions are manually approved. We always review changelogs provided by Depfu. **You can review updates any time [by filtering Merge Requests tagged `depfu`](https://gitlab.com/front-commerce/front-commerce/-/merge_requests?scope=all&state=merged&label_name[]=depfu).**
 
+### Storybook 6
+
+Your styleguide is now powered by Storybook 6. 
+
+This induces several changes in your custom stories.
+
+- The organization of stories is now only manageable with `/`, please set your story name like this : `section/path/of/your/story/Component`
+- As knobs are now deprecated, we highly encourages you to use controls from now on. see [the official documentation for more details](https://storybook.js.org/addons/@storybook/addon-knobs)
```
real34, line 30: Due to the impact, I think further references would be needed here. Examples:
- https://github.com/storybookjs/storybook/blob/next/addons/controls/README.md#how-do-i-migrate-from-addon-knobs
- https://storybook.js.org/docs/react/essentials/controls
- https://github.com/storybookjs/storybook/discussions/15060

Also a diff summarizing how to do it (from one of our updated component for instance) would make things clearer for FC.

Finally, I think we should mark it as deprecated in FC until 3.0 and document it here so that integrators can plan the migration on their side (it's not strictly needed when upgrading to 2.11)

source/docs/appendices/migration-guides.md
```diff
@@ -20,6 +20,16 @@ These new features may be relevant for your existing application:
...
+- The way to import markdown files in stories has changed, please move from `import readme from "./readme.md";` to `import { html as readme } from "./readme.md";`
```
real34, line 31: Now that we have codemods tooling in place. Wouldn't it be something quite fast to implement? So that developers would only have to run one command during the migration.

A `sed` command one-line could also makes the job here though…

I have no idea of how many instances we had in our codebase and if that was tedious or not.

Also, for such changes we usually use `diff` syntax to make it easier to scan and understand by developers (we got positive feedbacks for this practice):
```suggestion
- The way to import markdown files in stories has changed, please update your code as below:
```diff
- import readme from "./readme.md";
+ import { html as readme } from "./readme.md";
```
```

**Review by real34 — CHANGES_REQUESTED** [review]
I reviewed the migration guide before the actual MR (still in Draft).

Suggestions are more about making upgrade life easier for developers. If we can document automated ways to update stories, we should prefer this.\
It is better that we spend a bit more time on properly document an efficient way to do the change than every integrator doing it themselves.

Since we now have a base tooling for codemod, it may be better to create one if specific tasks requires a change across many files (if that is convenient for us, it makes our lives easier too).

There is also a storybook example not updated at https://github.com/front-commerce/developers.front-commerce.com/blob/115c1fc72b95f9439f4f608dbfbea3cc143072f5/source/docs/essentials/create-a-business-component.md?plain=1#L219

File: source/docs/appendices/migration-guides.md (hunk 1 above)
Comment (abailleulfc) [reply source/docs/appendices/migration-guides.md:30]: I did not write this doc as I don't understand myself how to do this migration and maintain everything working

File: source/docs/appendices/migration-guides.md [hunk 3]
```diff
@@ -20,6 +20,16 @@ These new features may be relevant for your existing application:
 - [A quickorder module has been created to order by SKU](/docs/advanced/features/quickorder.html)
 - Front-Commerce dependencies are now regularly (and automatically) updated using [Depfu](https://depfu.com/). Patches are automatically applied if the CI is green, minor versions are manually approved. We always review changelogs provided by Depfu. **You can review updates any time [by filtering Merge Requests tagged `depfu`](https://gitlab.com/front-commerce/front-commerce/-/merge_requests?scope=all&state=merged&label_name[]=depfu).**
 
+### Storybook 6
+
+Your styleguide is now powered by Storybook 6. 
+
+This induces several changes in your custom stories.
+
+- The organization of stories is now only manageable with `/`, please set your story name like this : `section/path/of/your/story/Component`
```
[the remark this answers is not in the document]
abailleulfc: The storybook codemod fails but I added this change to a FC codemod "storybook6". 
This codemod has been added to the migration guide.

source/docs/appendices/migration-guides.md
abailleulfc: import change added to a codemod (usage specified in the migration guide)

source/docs/appendices/migration-guides.md
```diff
@@ -20,6 +20,22 @@ These new features may be relevant for your existing application:
...
+- As knobs are now deprecated, we highly encourages you to use controls from now on. see [the official documentation for more details](https://storybook.js.org/addons/@storybook/addon-knobs)
+- The way to import markdown files in stories has changed, please update your code as below:
+```diff
+- import readme from "./readme.md";
++ import { html as readme } from "./readme.md";
+```
+
+Front-commerce provides a codemod to automatically migrate the hierarchy and readme imports. Run `npx jscodeshift -t scripts/codemod/transforms/storybook5-to-storybook6.js <module_path>` to make the necessary replacements in all .story.js files.
```
paul-vd, line 37: > Run `npx jscodeshift -t scripts/codemod/transforms/storybook5-to-storybook6.js <module_path>` to make the necessary replacements in all .story.js files

You can replace the `npx jscodeshift -t scripts/codemod/transforms/storybook5-to-storybook6.js <module_path>` with `front-commerce codemod -t storybook5-to-storybook6`, If the `storybook5-to-storybook6` codemod [exports the extensions](https://gitlab.com/front-commerce/front-commerce/-/blob/d0f09f27add2a6da50a7c427338deec0eee419d8/scripts/codemod/transforms/storybook5-to-storybook6.js#L67), then it will only run on `*..story.js` files. and it will scan folders based on the .front-commerce.js configuration.

paul-vd (changes requested):

source/docs/appendices/migration-guides.md
abailleulfc: command changed

real34:
Closing this PR as part of housekeeping.
To be rewritten when the Storybook upgrade will have landed in the core.

PR #257, opened 2021-11-04
chore: add migration docs for wp5

paul-vd:


real34:
Closing this staled Draft PR as part of housekeeping.
Webpack 5 will not land right away in FC, and might be reconsidered along with alternatives by the end of the year.

PR #271, opened 2021-12-03
FC search behavior/es pages indexing common mistake

mo-fc:
documented default FC search choices and the common mistake on M2 retated to page indexing

### DIFF
--- a/source/docs/magento1/search-engine.md
+++ b/source/docs/magento1/search-engine.md
@@ -7,7 +7,7 @@ As of Front-Commerce 2.5.0, it is possible to use Elasticsearch or Algolia to
 bring a search engine to your website.
 
 <blockquote class="warning">
-⚠️ NOTE: The default behaviour in Front-Commerce is that in the top search bar we search products, categories and pages(CMS Pages). While on the search page we search products only. This is an implementation choice we took. It is up to the integrator to modify these behaviour if needed (components and GraphQL queries under `theme/modules/Search` will need to be overriden).
+⚠️ NOTE: The default behavior in Front-Commerce is that in the top search bar we search products, categories and pages(CMS Pages). While on the search page we search products only. This is an implementation choice we took. It is up to the integrator to modify these behavior if needed (components and GraphQL queries under `theme/modules/Search` will need to be overriden).
 </blockquote>
 
 <br/>
--- a/source/docs/magento1/search-engine.md
+++ b/source/docs/magento1/search-engine.md
@@ -6,6 +6,12 @@ title: Search engine
 As of Front-Commerce 2.5.0, it is possible to use Elasticsearch or Algolia to
 bring a search engine to your website.
 
+<blockquote class="warning">
+⚠️ NOTE: The default behaviour in Front-Commerce is that in the top search bar we search products, categories and pages(CMS Pages). While on the search page we search products only. This is an implementation choice we took. It is up to the integrator to modify these behaviour if needed (components and GraphQL queries under `theme/modules/Search` will need to be overriden).
+</blockquote>
+
+<br/>
+
 ## Elasticsearch
 
 ### Requirements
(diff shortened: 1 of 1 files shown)


source/docs/magento1/search-engine.md
```diff
@@ -82,6 +90,146 @@ query Search {
...
+```
+diff --git a/src/web/theme/modules/Search/SearchBar/Autocomplete/Autocomplete.js b/src/web/theme/modules/Search/SearchBar/Autocomplete/Autocomplete.js
+index 9bc108484..ea9a2c91f 100644
+--- a/src/web/theme/modules/Search/SearchBar/Autocomplete/Autocomplete.js
++++ b/src/web/theme/modules/Search/SearchBar/Autocomplete/Autocomplete.js
+@@ -3,10 +3,9 @@ import PropTypes from "prop-types";
+ import { FormattedMessage } from "react-intl";
+ import ProductsResults from "./ProductsResults";
+ import CategoriesResults from "./CategoriesResults";
+-import PagesResults from "./PagesResults";
```
abailleulfc, line 112: suggestion not related to the doc

Wouldn't it be better to manage the full autocomplete search in a Loader server-side and return the result to the app ? 

This way an environment variable would be enough to disable some search features like CmsPages

abailleulfc (approved):

source/docs/magento1/search-engine.md
```diff
@@ -82,6 +90,146 @@ query Search {
 If you are using the default theme or the theme Chocolatine, the search bar
 should now be visible.
 
+<blockquote class="warning">
+⚠️ Common mistake: For magento 2 the elasticsuite does not index cms pages. To do that you need to install <a href="https://github.com/Smile-SA/magento2-module-elasticsuite-cms-search" target="_blank">Smile elasticsuite cms search module</a> on your magento instance. Failing to do so will cause the top search bar to malfunction. If you do not need the CMS Page search you can always override related `SearchBar` Components/GraphQL queries to disable the cms page search functionality as in the following diff:
```
dpobel, line 94: I think this is not correct. Either we state the CMS Search Magento module is a requirement then we document it as such (and then by definition, not following the requirements means a dysfunctioning platform), either we consider it as optional and in the case, FC should work properly without it.

Also here, you are modifying a Magento1 page mentioning a Magento2 module, this is very confusing. I know the corresponding page for Magento2 does not exist, it's a good opportunity to create it.

source/docs/magento1/search-engine.md
```diff
@@ -82,6 +90,146 @@ query Search {
...
+
+<details>
+<summary>
+Disable CMS Pages Search Diff
+</summary>
+
+#### Base theme:
+
+```
+diff --git a/src/web/theme/modules/Search/SearchBar/Autocomplete/Autocomplete.js b/src/web/theme/modules/Search/SearchBar/Autocomplete/Autocomplete.js
```
dpobel, line 104: this is a complete workaround to unlock the customer. IMHO, that should not be documented

dpobel (changes requested):

source/docs/magento1/search-engine.md
mo-fc: you need to update the client query anyways. so why search everything when not everything is needed.

the thing we should do imo is to ship the search bar with cms page disabled by default

source/docs/magento1/search-engine.md
mo-fc: > unlock the customer

what do you mean this?

mo-fc: this is in a toggleable item which means by default he cant see the diffs unless he clicks on them

dpobel: _unblock_ :)

even if it's toggleable, that's only a workaround either because the requirement is not properly documented or because there's a bug on our end. So let's fix the root cause and forget about it.

mo-fc:
closed cause of [this](https://github.com/front-commerce/developers.front-commerce.com/pull/271#discussion_r761928855) comment. we will fix the default behavior in FC. and then maybe create a documentation for M2 search and in that we can mention the cms pages and how to enable them

PR #275, opened 2021-12-10
fix: fixes url to config on root project

paul-vd:


real34 (changes requested):
Thanks for detecting this.\
While we're at it, could you please also fix all the other occurrences?

```
➜  developers.front-commerce.com git:(main) ✗ rg "develop/"
source/docs/reference/graphql-module-definition.md
94:  [`flattenAndReorderModulesUsingDependencies`](https://gitlab.com/front-commerce/front-commerce/blob/develop/src/server/core/graphql/__tests__/flattenAndReorderModulesUsingDependencies.js)

source/docs/reference/configurations.md
14:Each file described below exist in [`node_modules/front-commerce/src/config`](https://gitlab.com/front-commerce/front-commerce/tree/develop/src/config) in Front-Commerce. If you want to override some, duplicate those in `my-module/config`.

source/docs/advanced/theme/translations.md
52:    However, for this particular use case we have abstracted this in Front-Commerce with the [`theme/components/atoms/Price`](https://gitlab.com/front-commerce/front-commerce/tree/develop/src/web/theme/components/atoms/Typography/Price) component (and its variants: `ProductPrice`, `PriceOrFree`). We recommend you to use it instead, so you could benefit from better integration with its related GraphQL Type.

source/docs/advanced/server/add-http-endpoint.md
26:In the case of the WYSIWYG preview, let's say that when a user visits the `wysiwyg-preview` URL, we want to display static files that will let anyone preview how an html string will be transformed by [Front-Commerce's Wysiwyg component](https://gitlab.com/front-commerce/front-commerce/tree/develop/src/web/theme/modules/Wysiwyg).

source/docs/essentials/create-a-ui-component.md
22:and you could dive into the [`node_modules/front-commerce/src/web/theme/components`](https://gitlab.com/front-commerce/front-commerce/tree/develop/src/web/theme/components)
```

real34 (approved):
Thanks!

PR #293, opened 2022-01-10
fix: fixes code syntax in prismic docs

paul-vd:
Removes unclear braces in the arguments for the methods to query content.

```diff
- loadSingle(typeIdentifier[, options])
+ loadSingle(typeIdentifier, options)

- loadByUID(typeIdentifier, uid[, options])
+ loadByUID(typeIdentifier, uid, options)

- loadList(query[, options])
+ loadList(query, options) 
```

### DIFF
--- a/source/docs/prismic/expose-content.md
+++ b/source/docs/prismic/expose-content.md
@@ -15,10 +15,42 @@ To benefit from this API, you first need to [install the Prismic module](/docs/p
 
 The Prismic loader has the following API to request Content from Prismic:
 
-- `loadSingle(typeIdentifier[, options])` returns [an object of type `Content`](https://gitlab.com/front-commerce/front-commerce-prismic/-/blob/main/prismic/server/modules/prismic/core/loaders/Content.js) or throws an error if no such Content exists.
-- `loadByID(id[, options])` returns [an object of type `Content`](https://gitlab.com/front-commerce/front-commerce-prismic/-/blob/main/prismic/server/modules/prismic/core/loaders/Content.js) or throws an error if no such Content exists.
-- `loadByUID(typeIdentifier, uid[, options])` returns a `Content` representing a Prismic Content of [the corresponding type](https://prismic.io/docs/core-concepts/custom-types) and having [an UID field](https://prismic.io/docs/core-concepts/uid) with the given value. If such Content does not exist, it throws an error.
-- `loadList(query[, options])` returns [a `ContentList`](https://gitlab.com/front-commerce/front-commerce-prismic/-/blob/main/prismic/server/modules/prismic/core/loaders/ContentList.js) matching the query. `query` must be an instance of [`ListQuery`](https://gitlab.com/front-commerce/front-commerce-prismic/-/blob/main/prismic/server/modules/prismic/core/loaders/ListQuery.js), it provides a way to filter, sort and paginate Prismic Content.
+#### `loadSingle(typeIdentifier[, options])`
+
+Returns [an object of type `Content`](https://gitlab.com/front-commerce/front-commerce-prismic/-/blob/main/prismic/server/modules/prismic/core/loaders/Content.js) or throws an error if no such Content exists.
+
+**Arguments:**
+
+- typeIdentifier (string): The type of document.
+- options ([ContentTransformOptions](https://gitlab.com/front-commerce/front-commerce-prismic/-/blob/main/prismic/server/modules/prismic/core/loaders/index.js#L30-34)): Content transform options
+
+#### `loadByID(id[, options])`
+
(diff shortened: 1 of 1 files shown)


dpobel:
those brackets indicates that `options` parameter is optional. I think (but I might be wrong) that it's a quite popular convention in API or command documentation (unix man pages use that for instance) so if this is changed, we need a way to convey the same information.

paul-vd:
> I agree with @dpobel. We could start using Typescript-like definition to express the interface of a function / class / etc… We've tried it at https://developers.front-commerce.com/docs/reference/remote-schema-helpers.html#makeExecutor a bit, but it was never really pushed in the documentation so far.

I'm happy to leave as is, this was just to evoke readability of the docs and get some ideas.

Here are some examples for readability tests, I have also attached a more detailed example of api docs  at the end  which allows to better define different functionality.

_Current:_
```js
loadSingle(typeIdentifier[, options])
loadByUID(typeIdentifier, uid[, options])
loadList(query[, options])
```

_Typescript_
```js
loadSingle(typeIdentifier:string, options?:ContentTransformOptions)
loadByUID(typeIdentifier:string, uid:string, options?:ContentTransformOptions)
loadList(typeIdentifier:string, options?:ContentTransformOptions)
```
- [ContentTransformOptions](https://gitlab.com/front-commerce/front-commerce-prismic/-/blob/main/prismic/server/modules/prismic/core/loaders/index.js#L30-34)

_Proposed:_
```js
loadSingle(typeIdentifier, options)
loadByUID(typeIdentifier, uid, options)
loadList(query, options) 
```
_A More Defined Example_

> The Prismic loader has the following API to request Content from Prismic:
>
> ### `loadSingle(typeIdentifier, options)`
> Returns [an object of type `Content`](https://gitlab.com/front-commerce/front-commerce-prismic/-/blob/main/prismic/server/modules/prismic/core/loaders/Content.js) or throws an error if no such Content exists.
> #### Arguments:
> - typeIdentifier (string): The type of document.
> - [options] ([ContentTransformOptions](https://gitlab.com/front-commerce/front-commerce-prismic/-/blob/main/prismic/server/modules/prismic/core/loaders/index.js#L30-34)): Content transform options 
> 
> ### `loadByUID(typeIdentifier, uid, options)` 
> Returns a `Content` representing a Prismic Content of [the corresponding type](https://prismic.io/docs/core-concepts/custom-types) and having [an UID field](https://prismic.io/docs/core-concepts/uid) with the given value. If such Content does not exist, it throws an error.
>  #### Arguments:
> - `typeIdentifier (string)`: The type of document.
> - uid (string): The Unique ID of the document.
> - [options] ([ContentTransformOptions](https://gitlab.com/front-commerce/front-commerce-prismic/-/blob/main/prismic/server/modules/prismic/core/loaders/index.js#L30-34)): Content transform options 
> 
> ### `loadList(query, options)`
> Returns [a `ContentList`](https://gitlab.com/front-commerce/front-commerce-prismic/-/blob/main/prismic/server/modules/prismic/core/loaders/ContentList.js) matching the query. `query` must be an instance of [`ListQuery`](https://gitlab.com/front-commerce/front-commerce-prismic/-/blob/main/prismic/server/modules/prismic/core/loaders/ListQuery.js), it provides a way to filter, sort and paginate Prismic Content.
> #### Arguments:
> - query ([ListQuery](https://gitlab.com/front-commerce/front-commerce-prismic/-/blob/main/prismic/server/modules/prismic/core/loaders/ListQuery.js)): A query for a list of documents
> - [options] ([ContentTransformOptions](https://gitlab.com/front-commerce/front-commerce-prismic/-/blob/main/prismic/server/modules/prismic/core/loaders/index.js#L30-34)): Content transform options

source/docs/prismic/expose-content.md
```diff
@@ -15,9 +15,9 @@ To benefit from this API, you first need to [install the Prismic module](/docs/p
 
 The Prismic loader has the following API to request Content from Prismic:
 
-* `loadSingle(typeIdentifier[, options])` returns [an object of type `Content`](https://gitlab.com/front-commerce/front-commerce-prismic/-/blob/main/prismic/server/modules/prismic/core/loaders/Content.js) or throws an error if no such Content exists.
```
abailleulfc, line 18: Why is it unclear ? Aren't "options" optional ?

real34:
@paul-vd all those suggestions looks good, I have no strong opinions on those but have a small preference for the long version. At some point I think that they'll have to automatically be extracted from the code BTW.

mo-fc:
the longer version looks cool. but you are using the `[]` notation to indicate that they are optional, which is the reason of the PR in the first place. maybe add a note that they are optional between brackets

real34 (approved):

PR #295, opened 2022-01-12
Update expose-content.md

Aerowiel:
If we assign the result of the loadList to a variable I think we need to await it before returning it.


paul-vd (approved):
Thank you for your contribution!

PR #301, opened 2022-01-13
Various fixes: indent, numbering, …

dpobel:
noticed all over the place while browsing the documentation.

### DIFF
--- a/source/docs/essentials/extend-the-theme.md
+++ b/source/docs/essentials/extend-the-theme.md
@@ -81,9 +81,8 @@ You will thus need to update the fragment related to `ProductItem`.
 The original fragment file is collocated with the original component at: [`node_modules/front-commerce/src/web/theme/modules/ProductView/ProductItem/ProductItemFragment.gql`](https://gitlab.com/front-commerce/front-commerce/blob/main/src/web/theme/modules/ProductView/ProductItem/ProductItemFragment.gql)
 
 1. copy it to: `my-module/web/theme/modules/ProductView/ProductItem/ProductItemFragment.gql`
-2. add the field `description` to the fragment.
-
-```diff
+1. add the field `description` to the fragment.
+  ```diff
 // my-module/web/theme/modules/ProductView/ProductItem/ProductItemFragment.gql
  fragment ProductItemFragment on Product {
    path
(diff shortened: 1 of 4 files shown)


source/docs/essentials/extend-the-theme.md
```diff
@@ -81,9 +81,8 @@ You will thus need to update the fragment related to `ProductItem`.
 The original fragment file is collocated with the original component at: [`node_modules/front-commerce/src/web/theme/modules/ProductView/ProductItem/ProductItemFragment.gql`](https://gitlab.com/front-commerce/front-commerce/blob/main/src/web/theme/modules/ProductView/ProductItem/ProductItemFragment.gql)
 
 1. copy it to: `my-module/web/theme/modules/ProductView/ProductItem/ProductItemFragment.gql`
-2. add the field `description` to the fragment.
-
-```diff
+1. add the field `description` to the fragment.
```
abailleulfc, line 84: We already have a "1" at line 83, shouldn't it be a "2" ?

```suggestion
2. add the field `description` to the fragment.
```

source/docs/essentials/extend-the-theme.md
```diff
@@ -92,9 +91,8 @@ The original fragment file is collocated with the original component at: [`node_
    ...ProductPriceFragment
    imageUrl
    sku
-```
-
-3. restart the application
+  ```
+1. restart the application
```
abailleulfc, line 95: Here is a "1" again after two other "first point", the "3" looked good 

```suggestion
3. restart the application
```

source/docs/prismic/routable-types.md
```diff
@@ -179,51 +177,48 @@ Album: {
...
++ changefreq: () => "daily",
 },
 };
-```
-
-2. Enable sitemap in the module definition file:
-
-```diff
+  ```
+1. Enable sitemap in the module definition file:
```
abailleulfc, line 187: Why did you change for several "first point" in a list of steps ?

abailleulfc (changes requested):

source/docs/prismic/routable-types.md
dpobel: so that the markdown parser automatically adds the right number and the resulting HTML markup is a proper numbered list, see https://deploy-preview-301--heuristic-almeida-1a1f35.netlify.app/docs/prismic/routable-types.html#Add-a-routable-custom-type-to-the-sitemap for this example

Also, if we have to reorder/insert/remove a list item, you don't have to worry about the number.

abailleulfc: Ok, I didn't know that it worked that way

abailleulfc (approved):

real34 (approved):

PR #312, opened 2022-01-28
Adds example for registerPrismicRoute method

paul-vd:

PR #313, opened 2022-02-01
feat(prismic): add example for path on routeable types

paul-vd:


source/docs/prismic/routable-types.md
```diff
@@ -69,7 +69,9 @@ Now you need to add your type to the GraphQL dispatcher query. For more info on
 
 ## Register the Prismic custom type as a routable type
 
-To make a custom Prismic type routable it must be registered with the Prismic module. First ensure that the `Prismic/Core` is added as a module's dependency then use `loaders.Prismic.registerRoutableType` to add the custom type as a routable type:
+To make a custom Prismic type routable it must be registered with the Prismic module. First ensure that the `Prismic/Core` is added as a module's dependency then;
+- use `loaders.Prismic.registerRoutableType` to add the custom type as a routable type
+- use `loaders.Prismic.registerPrismicRoute` to add allow the custom type to be handled by Prismic, see [Route Resolver](https://prismic.io/docs/core-concepts/link-resolver-route-resolver#route-resolver) for more information
```
real34, line 74: **issue:** it is not clear about the differences. Readers may think both calls are needed

**suggestion:** we should make each use case a bit clearer to better match the use case encountered by the developer

```suggestion
- either use `loaders.Prismic.registerRoutableType` to add the custom type as a routable type
- or use `loaders.Prismic.registerPrismicRoute` to inform Prismic about a custom URL pattern for a custom type, see [Route Resolver](https://prismic.io/docs/core-concepts/link-resolver-route-resolver#route-resolver) for more information
```

source/docs/prismic/routable-types.md
```diff
@@ -109,8 +113,21 @@ export default {
 };
 ```
 
-Note the `postTransformer` above is optional. It is a function that will be called after [the transformation is done using `contentTransformOptions`](/docs/prismic/expose-content.html#Field-Transformers). It is given the current URL being resolved and the transformed document. It can be used if you have some custom logic to apply to the document or if you want to prevent the document from showing using some custom logic (returning a _falsy_ value).
+<strong>Note:</strong> the `postTransformer` above is optional. It is a function that will be called after [the transformation is done using `contentTransformOptions`](/docs/prismic/expose-content.html#Field-Transformers). It is given the current URL being resolved and the transformed document. It can be used if you have some custom logic to apply to the document or if you want to prevent the document from showing using some custom logic (returning a _falsy_ value).
```
real34, line 116:
```suggestion
**Note:** the `postTransformer` above is optional. It is a function that will be called after [the transformation is done using `contentTransformOptions`](/docs/prismic/expose-content.html#Field-Transformers). It is given the current URL being resolved and the transformed document. It can be used if you have some custom logic to apply to the document or if you want to prevent the document from showing using some custom logic (returning a _falsy_ value).
```

source/docs/prismic/routable-types.md
```diff
@@ -109,8 +113,21 @@ export default {
...
 
+<strong>Note:</strong>
```
real34, line 118:
```suggestion
**Note:**
```

real34 (approved):

source/docs/prismic/routable-types.md
paul-vd: Sorry yes, this is still a remenant of https://github.com/front-commerce/developers.front-commerce.com/pull/312 , I missed this, I will remove it for now, because I wan't to better document this under `Previews` where this becomes relevant.

PR #334, opened 2022-02-21
chore(configuration): Updated documentation upon removal of default CSPs in Front-Commerce

Kagaminara:
- Updated CSPs configuration in Essentials' `Display a map` documentation
- Updated CSPs configuration in `Analytics`, `Paypal` and `Payzen` modules documentation
- Initialized migration guide to `2.14.0`

In the Analytics documentation, I had to not indent the `diff`s code blocks because it breaks the color highlighting, so I also removed the indentations on the near code blocks for consistency. 

_Note: There's quite a few changes that are due to my Prettier, should I disable it?_


source/docs/advanced/theme/analytics.md
```diff
@@ -385,7 +412,7 @@ PixelIntegration.prototype.event = function (track) {
     img.setAttribute("src", pixelUrl);
     img.setAttribute(
       "style",
-      "position: absolute; left: -5000px; top: -4000px;"
+      "position: absolute; left: -5000px; top: -4000px;",
```
abailleulfc, line 415: This is a strange autoformat as we do not use this rule in FC itself. 
Was it from the a formatting config inside the project or a tool you have ?

Kagaminara: I think it comes from my VSCode's Prettier addon. I though it used the config found in the opened folder, but it seems it has a default config somewhere.
Do we use a specific formatter at FC (if any) ?

Kagaminara: @abailleulfc Suggestion: Should we be really exhaustive in detailing prettier rules in `.prettierrc`, so that the IDEs don't override them?

Kagaminara: Btw, I just spotted `"trailingComma": "all"` in the `.prettierrc`, so I guess we do indeed have this rule?

abailleulfc (approved):

source/docs/advanced/theme/analytics.md
abailleulfc: Then for this PR it looks good to me (moving the topic to slack)

real34:
The target branch was mistakenly defined to `main`.
A consequence is that it is now live at https://developers.front-commerce.com/docs/appendices/migration-guides.html

Since this documentation is for things related to FC 2.14, could you please revert the commit in `main` and cherry-pick only in `next`?
The consequence is that it will appear on https://developers.next.front-commerce.com/docs/appendices/migration-guides.html and not on https://developers.front-commerce.com/docs/appendices/migration-guides.html

source/docs/advanced/theme/analytics.md
real34: Config updated in #342. I'm marking this conversation as resolved.

PR #340, opened 2022-02-25
ci(lint): add a `npm run lint` script to validate our style conventions

real34:
This PR will serve for an initial Github action setup with the linting.


real34:
Closed in favor of #341 so Github actions can detect the newly added workflow.

PR #346, opened 2022-03-02
feat(dx): add relative paths for search result

paul-vd:
### Why?
If you search in any instance other than production, the url's returned are absolute (with production origin), which results in unusable searching in either local development or preview deployments, this addresses that issue by removing the origin from the URL, only exposing the pathname, which results in navigatable links in current deployment/instance.

| Berfore | After    |
| ---------- | --------- |
|  <video src="https://user-images.githubusercontent.com/39598117/156334515-abd1aebf-747e-4550-8603-78b0040a39bb.mp4" />  |   <video src="https://user-images.githubusercontent.com/39598117/156334583-dd6ab3d4-b4cb-4a1d-a0c4-a46142ac8ab5.mp4" />  |

### DIFF
--- a/theme/layout/page.ejs
+++ b/theme/layout/page.ejs
@@ -63,22 +63,22 @@
             }
           %>
           <% if (prev) { %>
-         
+
               <a class="link primary prev" href="<%- relative_url(page.path, prev.path) %>">
                 <span class="icon-arrow-left-alt"></span>
                 <span class="subtitle-pagination">Previous</span>
                 <%- prev.title %>
               </a>
-         
+
           <% } %>
           <% if (next) { %>
-           
+
               <a class="link primary next" href="<%- relative_url(page.path, next.path) %>">
                 <span class="subtitle-pagination">Next</span>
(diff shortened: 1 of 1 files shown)


paul-vd:
One issue I can see here is dead links if the link has been removed in the current version but still available on production. because it will still be indexed.

But this should not be such big of a deal as it's mostly development facing.

dpobel (approved):
Huge :+1: this has been a painful issue when working on next documentation or even in preview generated for each MR!

abailleulfc (approved):

real34:
That's awesome!

Would you be interested in upgrading the docsearch experience too?
I have a TODO in my emails since months and never took the time to dig into this… that'd also make search even better I think.

paul-vd:
> That's awesome!
> 
> Would you be interested in upgrading the docsearch experience too? I have a TODO in my emails since months and never took the time to dig into this… that'd also make search even better I think.
> 
> I'll forward the email to the team!

@real34 Sure

PR #355, opened 2022-03-10
Give a complete translation without React component example

dpobel:
`defineMessages` is missing in the example.


paul-vd (approved):
Should we not also include inline examples, because now we are only showing an example with messages?

dpobel:
> Should we not also include inline examples, because now we are only showing an example with messages?

actually that's the point of this change if you use inline message, npm run translate can not extract messages so it's better to not document that.

PR #359, opened 2022-03-14
Fc 822 cache strategy per country

abailleulfc:


source/docs/advanced/graphql/dataloaders-and-cache-invalidation.md
```diff
@@ -420,6 +420,51 @@ export default {
 };
 ```
 
+### `PerMagentoCustomerTaxZone`
+
+_Since version 2.14.0_
+</blockquote>
+
+The `PerMagentoCustomerTaxZone` implementation is a decorator that is specific to Magento 1 and Magento 2 modules. It will decorate the existing caching strategies so that DataLoader keys are specific to the current customer's tax zone. We highly recommend to use it on Magento stores that have price with taxes depending on several tax zones, so they can leverage other caching mechanisms (such as `Redis`).
```
dpobel, line 429: it's only for Magento2, no ?

source/docs/advanced/graphql/dataloaders-and-cache-invalidation.md
```diff
@@ -420,6 +420,51 @@ export default {
...
+
+Has such strategies can be complex, the tax zone definition is let to the integrator through the `taxZoneKeyFromAddress` function
```
dpobel, line 431:
```suggestion
As such strategy can be complex, the tax zone definition is left to the integrator through the `taxZoneKeyFromAddress` function
```

source/docs/advanced/graphql/dataloaders-and-cache-invalidation.md
```diff
@@ -420,6 +420,51 @@ export default {
...
+        "CatalogPrice",
+        "CatalogProductChildrenPrice",
+        "CatalogProductBundlePrice",
+      ],
+      config: {
+        addressType: "shipping", // or "billing" depending on which address is used to define the taxes showed to the user
+        defaultTaxZoneKey: "FR",
+        taxZoneKeyFromAddress: (address) => {
+          // see https://docs.magento.com/user-guide/tax/tax-zones-rates.html
+          return; // default implementation uses the defaultTaxZoneKey value
```
dpobel, line 460: is it intended to return nothing ? that looks weird (unfinished) from an outside point of view ?

source/docs/advanced/graphql/dataloaders-and-cache-invalidation.md
abailleulfc: indeed it is only Magento2, changed

dpobel (approved):

PR #363, opened 2022-03-21
chore(prismic): add advanced path usage examples

paul-vd:


source/docs/prismic/routable-types.md
```diff
@@ -347,3 +336,38 @@ loaders.Prismic.registerPrismicRoute({
 ```
 
 > This method will not create a resolvable route in Front-Commerce. If you need a routable type please use the `registerRoutableType` method instead, it also registers a prismic route.
+
+## Advanced usage
+
+### Trailing Slash
+
+By default Front-Commerce will redirect urls with trailing slashes to their counterpart without a trailing slash. For example `/album/` will redirect to `/album`. You can configure this behavior to act the opposite way, where urls without trailing slashes are redirected to their counterparts with trailing slashes, you can simply add a trailing slash to the `path` on the `registerRoutableType` method.
```
Kagaminara, line 344: Nit pick, but I find the second sentense a bit weird (although I can really find a better way to right it).
"You can [...], where [...], you can [...]".
Could a "If you want to ..." work better? I don't know.

Kagaminara (approved):

source/docs/prismic/routable-types.md
paul-vd: I separated the description and the usage in two paragraphs, it should read better now?

Kagaminara (approved):

PR #365, opened 2022-03-23
Document Payment method label translation key change

mo-fc:


Kagaminara (approved):
Lgtm !

real34 (changes requested):
It is not clear for integrators that the PaymentLabel component has been relocated too.

Changing translation keys is wrong if they've overriden it. I think we must document the component relocation and mention the translation key update as a consequence.

source/docs/appendices/migration-guides.md
```diff
@@ -9,6 +9,26 @@ Our goal is to make migrations as smooth as possible. This is why we try to make
 
 ## `2.13.0` -> `2.14.0`
 
+### PaymentMethodLabel relocated
+
+The `<PaymentMethodLabel>` component was moved from `theme/modules/User/Order/OrderMethod/PaymentMethodLabel.js` to `theme/modules/Checkout/Payment/PaymentMethodLabel.js` the old location will still work but will output deprecation messages when used. If you have overrided `<PaymentMethodLabel>` you should also relocate your override to the same path and update all references to point to it.
```
abailleulfc, line 14:
```suggestion
The `<PaymentMethodLabel>` component was moved from `theme/modules/User/Order/OrderMethod/PaymentMethodLabel.js` to `theme/modules/Checkout/Payment/PaymentMethodLabel.js` the old location will still work but will output deprecation messages when used. If you have overriden `<PaymentMethodLabel>` you should also relocate your override to the same path and update all references to point to it.
```

abailleulfc (changes requested):

real34 (approved):

PR #384, opened 2022-04-12
chore: fix 'the the' grammar

paul-vd:
If not why not? https://github.com/front-commerce/developers.front-commerce.com/pull/377#discussion_r848534596 😁 happy to see I'm not 'the the' only one


real34 (approved):
Thanks! :smile:

PR #396, opened 2022-05-03
fix(config): clarify the use of the httpAuth configuration

abailleulfc:
# Why

Improved the httpAuth documentation to avoid misunderstanding the way this configuration works as it is not obvious

https://deploy-preview-396--heuristic-almeida-1a1f35.netlify.app/docs/reference/configurations.html#config-httpAuth-js


Kagaminara (approved):

mo-fc:
mmm this exposes the fact that these headers configurations are placed in the wrong place :(

real34 (approved):
It at least makes things clear until this is refactored in a clean config provider

PR #397, opened 2022-05-04
Document a minor breaking change in `dotenv` upgrade

real34:
As [per our discussion](https://gitlab.com/front-commerce/front-commerce/-/merge_requests/1231#note_927541557), this PR documents possible issues integrators could have in their project when upgrading to 2.15.

:arrow_right: https://deploy-preview-397--heuristic-almeida-1a1f35.netlify.app/docs/appendices/migration-guides.html#Environment-variables-quote-values-containing-or-characters

### DIFF
--- a/source/docs/appendices/migration-guides.md
+++ b/source/docs/appendices/migration-guides.md
@@ -9,7 +9,28 @@ Our goal is to make migrations as smooth as possible. This is why we try to make
 
 ## `2.14.0` -> `2.15.0`
 
-Nothing yet!
+### Environment variables: quote values containing `#` or ``` characters
+
+As part of our continuous dependencies upgrade process, we've upgraded the `dotenv` dependency to its latest version.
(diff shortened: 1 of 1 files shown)


abailleulfc:
I presume quoting those values for older Front-Commerce versions causes no trouble ? 

In this case I would clarify this so everyone knows it is possible to release this migration step without downtimes.

real34:
> I presume quoting those values for older Front-Commerce versions causes no trouble ?
> 
> In this case I would clarify this so everyone knows it is possible to release this migration step without downtimes.

That is a good idea :+1: 
I amended the PR with a blockquote mentioning this.

> Note: you can change your environment variables right away, as quoted values will also work with previous Front-Commerce versions.

abailleulfc (approved):

paul-vd (approved):

PR #403, opened 2022-05-12
Document a known limitation with ElasticSuite for native M2 categories

real34:
This MR adds a section to the Troubleshooting area about https://gitlab.com/front-commerce/front-commerce/-/merge_requests/1264#note_944172133

Preview: https://deploy-preview-403--heuristic-almeida-1a1f35.netlify.app/docs/appendices/troubleshooting.html#I-have-an-error-Cannot-read-property-39-label-39-of-null-with-Magento-2-native-categories-listing


dpobel (approved):
(I approve even if I think we should be way more precise in our requirements but we can take that later)

PR #411, opened 2022-06-15
chore: update package.lock to version 2

paul-vd:
As we have started using the latest npm for the core, it also makes sense to propagate that to the docs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,9936 @@
 {
   "name": "hexo-site",
   "version": "0.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -1,8 +1,9936 @@
...
+      "dev": true,
+      "optional": true,
+      "engines": {
+        "node": "*"
+      }
+    },
+    "node_modules/form-data2": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/form-data2/-/form-data2-1.0.3.tgz",
+      "integrity": "sha1-y6XiNgGmlE2Vq31xEf+Tl6XLKk0=",
```
real34, line 2142: small typo here, incorrect version used.

PR #414, opened 2022-06-20
Disable hydration

mo-fc:


mo-fc:
[Preview](https://deploy-preview-414--heuristic-almeida-1a1f35.netlify.app/docs/advanced/features/disable-hyration.html)

_config.yml
```diff
@@ -68,6 +68,9 @@ sidebar_categories:
     - docs/advanced/features/external-logins
     - docs/advanced/features/return-merchandise-authorization
     - docs/advanced/features/in-stock-alert
+    - docs/advanced/features/maintenance-mode
+    - docs/advanced/features/disable-hyration
```
real34, line 72: A link to this page should also be added to https://developers.front-commerce.com/docs/advanced/theme/server-side-rendering.html#Learn-more for discoverability

source/docs/advanced/features/disable-hyration.md
```diff
@@ -0,0 +1,33 @@
+---
+id: disable-hyration
+title: Disable Hydration
+---
+
+# What is Hydration?
```
real34, line 6: An introduction is missing for context of this page. As per !408, it should be splitted into a `description:` and introduction paragraph.

If that feels like too much to do, then this page could also be reworked as a paragraph in the SSR page.

source/docs/advanced/features/disable-hyration.md
```diff
@@ -0,0 +1,33 @@
...
+```
+<script>
+  window.__HYDRATE__ = (function() {
+    var shouldHydrate = some_logic_to_determine_if_page_should_hydrate;
+    return shouldHydrate;
+  })();
+</script>
+```
+
+A typical logic to use to determine if one should hydrate or not is to inspect the `navigator.userAgent` variable to determine for example if it is the Google crawler bot.
```
real34, line 29: **issue:** I'm not sure this recommendation is relevant as we don't have any clue of its impact (it can lead to cloaking).

**suggestion:** we could illustrate with the example we got, to prevent hydration on different domain.\
```
window.__HYDRATE__ = location.host !== "webcache.googleusercontent.com";
```

real34 (changes requested):

source/docs/advanced/features/disable-hyration.md
mo-fc: I added a discription in similar manner to !408 the introduction remains the same. is that satisfactory?

PR #457, opened 2022-09-05
fix: missing images on various pages

dpobel:
from support

visible at: https://deploy-preview-457--heuristic-almeida-1a1f35.netlify.app/docs/advanced/theme/analytics#google-tag-manager (bottom of the page) vs. https://developers.front-commerce.com/docs/advanced/theme/analytics#google-tag-manager

### DIFF
--- a/docs/advanced/theme/analytics.mdx
+++ b/docs/advanced/theme/analytics.mdx
@@ -639,16 +639,18 @@ First, the `userConsents` configuration option will be pushed to your dataLayer
 as the `userConsents` value. You can reference it from a Variable in GTM. Here
 is an example:
 
-<Figure
-  src={require("./assets/gtm-datalayer-variable.png").default}
-  alt="Screenshot of a GTM Variable configured to expose the user consents"
-/>
+<Figure>
+
+![Screenshot of a GTM Variable configured to expose the user consents](./assets/gtm-datalayer-variable.png)
+
+</Figure>
 
 Then, you can leverage the `UserConsentUpdated` event tracked whenever users
 update their consent preferences. You could create triggers to enable scripts to
 load / remove (depending on the `userConsents` value). Here is an example:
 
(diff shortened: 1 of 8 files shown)


docs/advanced/theme/analytics.mdx
```diff
@@ -639,16 +639,10 @@ First, the `userConsents` configuration option will be pushed to your dataLayer
 as the `userConsents` value. You can reference it from a Variable in GTM. Here
 is an example:
 
-<Figure
-  src={require("./assets/gtm-datalayer-variable.png").default}
-  alt="Screenshot of a GTM Variable configured to expose the user consents"
-/>
+![Screenshot of a GTM Variable configured to expose the user consents](./assets/gtm-datalayer-variable.png)
```
Kagaminara, line 642: Isn't `<Figure>` the intended way of including images?

paul-vd: Using the `<Figure>` component will allow images to display the same across all docs, although it's not a requirement, if I remember correctly I centered figure images and also added some y-axis padding.

real34: The question then is: why was it broken here?

It seems that the `require()` path was incorrect and changing with a raw md image tag makes it work.

**issue:** we may have other broken images all over the documentation that would have to be fixed too

**suggestion:** fix the `<Figure>` usage instead, so that it works all over the doc.

dpobel: yeah I've discovered there are several others broken images of the same kind. And to me the `<Figure>` API is weird, it accepts both a children and/or a `src` props while IMHO, the markdown syntax is much better and `<Figure>` should probably only be a decorator for that.

So I'm gonna fix broken images and refactor Figure

real34 (approved):

PR #458, opened 2022-09-05
fix(redirects): Removed trailing slashes in redirect urls

Kagaminara:


paul-vd:
This is an attempt to fix broken rewrites and redirects after updating the netlify config to exclude backslashs, see https://docusaurus.io/docs/deployment#deploying-to-netlify

image

paul-vd (approved):
At some stage we need to review which urls should be permanent and which should be temporary. 

For safety we keeping all as temporary, but this could affect SEO in the future.

paul-vd:
Previously we used rewrites for overview pages, but the rewrites do not support [placeholders](https://docs.netlify.com/routing/redirects/redirect-options/#placeholders) in urls, so we opted to rather use redirects, which is technically method for these paths.

PR #460, opened 2022-09-06
Improve PageBuilder images

paul-vd:


dpobel (commented):
speaking of images, this PR adds 3 big images of about 1,5Mb each. Given, they are mainly composed of photos, I would suggest to use jpg instead of png or if you keep png, I'm pretty sure they can be optimized (and they probably don't kneed to be that big)

docs/magento2/page-builder.mdx
```diff
@@ -221,3 +222,170 @@ export default {
...
+> _Since version 2.18_
+
+The `image` content-type uses defaults for the formatting to optimise
+performance,
+
+- Desktop: `large`
+- Mobile: `medium`
+
+This might be limiting for content creators as they might want to display
+dynamicically sized images, or use a different format.
```
abailleulfc, line 237: hicup :smile: 
```suggestion
dynamically sized images, or use a different format.
```

docs/magento2/page-builder.mdx
```diff
@@ -221,3 +222,170 @@ export default {
...
+
+You can register your own image content type, and use a custom format, or you
+can determine the format dynamicically using a class.
```
abailleulfc, line 240: drink some water :wink: 
```suggestion
can determine the format dynamically using a class.
```

docs/magento2/page-builder.mdx
```diff
@@ -221,3 +222,170 @@ export default {
...
+
+### Using a custom format
+
+This is much simpler as we only need to register a Image content-type with the
```
abailleulfc, line 244:
```suggestion
This is much simpler as we only need to register an Image content-type with the
```

abailleulfc (changes requested):

docs/magento2/page-builder.mdx
paul-vd: 🙉 🙈 🙊

paul-vd:
> speaking of images, this PR adds 3 big images of about 1,5Mb each. Given, they are mainly composed of photos, I would suggest to use jpg instead of png or if you keep png, I'm pretty sure they can be optimized (and they probably don't need to be that large)

@dpobel I have updated the images

dpobel (approved):

dpobel:
looks like all remarks have been fixed, I'm merging.

PR #470, opened 2022-09-14
Improve discoverability for a native M2 search known issue with

real34:
ElasticSuite

so that users (and FC support 😇) can find it more easily

Preview: /docs/magento2/search-engine#native-search


dpobel (approved):

real34:
I merged `main` back into `next` to make this change available on both branches.

PR #471, opened 2022-09-15
docs(express): Documented way to override express configuration provider

Kagaminara:
This aims to help integrator override express configuration by providing an easy copy-pastable example in their code.

### DIFF
--- a/docs/advanced/server/configurations.mdx
+++ b/docs/advanced/server/configurations.mdx
@@ -366,6 +366,35 @@ const serviceOverrideProvider = {
 configService.insertAfter("serviceProvider", serviceOverrideProvider);
 ```
 
+### Overriding express configuration provider
+
+If you want to override the configuration of express, you will need to add the
+new configuration in a middleware definition file `module.config.js`.
+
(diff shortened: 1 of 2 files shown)


Kagaminara:
[Preview ](https://deploy-preview-471--heuristic-almeida-1a1f35.netlify.app/docs/advanced/server/configurations#overriding-express-configuration-provider)

real34 (approved):
:+1: 

For the record, express configuration provider is a bit special because of a chicken-and-egg problem:
- overriden values must be injected in other middlewares
- it means that they must be in `staticConfigFromProviders`
- but the `module.config.js` is imported at the beginning of the server entry (`express/index.js`), which  evaluates before the `expressConfigProvider` is appended to the `configService`

Those are the reasons why the documented solution is:
- in `module.config.js`
- overriding values using `values` (they will override the `default` values defined in the `expressConfigProvider` provided `schema`)
- but registered using `configService.append` instead of a `configService.insertAfter`

That is the reason why this needs to be documented _as-is_ in a copy-pastable way (no need to document these implementation details that might change in the future anyway)

PR #484, opened 2022-09-26
fix(cloud): caching and session cloud variables

abailleulfc:
# What

Following support discussions, the need to clarify the config files for cloud appeared.

I took this opportunity to introduce the new FRONT_COMMERCE_CLOUD_REDIS_DB which will need to be implemented at some point.

# Links

- https://deploy-preview-484--heuristic-almeida-1a1f35.netlify.app/docs/reference/configurations#configcachingjs
- https://deploy-preview-484--heuristic-almeida-1a1f35.netlify.app/docs/advanced/graphql/dataloaders-and-cache-invalidation#redis
- https://deploy-preview-484--heuristic-almeida-1a1f35.netlify.app/docs/advanced/production-ready/sessions

### DIFF
--- a/docs/advanced/graphql/dataloaders-and-cache-invalidation.mdx
+++ b/docs/advanced/graphql/dataloaders-and-cache-invalidation.mdx
@@ -458,7 +458,7 @@ It uses node's [`redis` package](https://www.npmjs.com/package/redis) as a redis
 client. Please refer to the package documentation for
 [all the available options](https://www.npmjs.com/package/redis#options-object-properties).
 
-Here is a configuration example:
+Here is a full configuration example:
 
 ```js title="my-module/config/caching.js"
 export default {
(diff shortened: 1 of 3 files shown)


docs/reference/configurations.mdx
```diff
@@ -262,7 +264,7 @@ export default {
       supports: "*",
       config: {
         host: "redis",
-        // host: "127.0.0.1"
+        // full configuration at https://developers.front-commerce.com/docs/advanced/graphql/dataloaders-and-cache-invalidation#redis
```
abailleulfc, line 267: I hesitated to paste here the full configuration too :thinking: any opinion ?

docs/advanced/graphql/dataloaders-and-cache-invalidation.mdx
```diff
@@ -468,9 +468,18 @@ export default {
       supports: "*",
       config: {
         // see https://www.npmjs.com/package/redis#options-object-properties
-        host: "127.0.0.1",
-        port: 6379,
-        db: 1,
+        host:
+          process.env.FRONT_COMMERCE_CLOUD_REDIS_HOST || // Front-Commerce Cloud variables
```
dpobel, line 472: instead of mixing the documentation of the cache with the cloud documentation, wouldn't it make sense to add a page dedicated to the cloud setup ?

docs/reference/configurations.mdx
Kagaminara: Maybe put it as a link inside a  `:::info` block below the code snipet ? I think it might make it more visible than inside a comment (if the redis part is the only important part here).

Kagaminara (approved):

docs/reference/configurations.mdx
abailleulfc: I would say it becomes a bit too visible in an info (this current setting will nearly never work as is for the developer, leading to follow the link in the code block to configure it properly)

PR #494, opened 2022-10-03
fix: encoding of adminData local testing

paul-vd:
### Why?

We parse the data on the query from a encoded string, this previous example would cause the parsed data to result in  `"foo" : "bar"` instead of `{ foo: "bar" }` this would result in a syntaxt error

```
"foo":"bar"
SERVER  500 Unexpected server error
SyntaxError: Unexpected token : in JSON at position 5
```

I split them up in two commands (easier more readable running in browser, and second encoded to run in curl)

https://user-images.githubusercontent.com/39598117/194026292-fece7476-4a05-4228-ace0-bc88b0e6eee5.mp4

### DIFF
--- a/docs/magento2/detect-admin-users.mdx
+++ b/docs/magento2/detect-admin-users.mdx
@@ -9,6 +9,9 @@ description:
   how Front-Commerce supports these use cases.
 ---
 
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
 <p>{frontMatter.description}</p>
 
 > _This feature has been added in version `2.1.0`_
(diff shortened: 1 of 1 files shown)


docs/magento2/detect-admin-users.mdx
```diff
@@ -238,7 +238,7 @@ without worrying about session cookies issues.
 Here is a sample `curl` request to achieve this:
```
abailleulfc, line 238: Maybe highlight the need to escape the JSON body ? 

```suggestion
Here is a sample `curl` request to achieve this (note the escaped JSON data) :
```

paul-vd: Please have a look at the video in the description, I split the two commands up, as @real34 showed me this morning that it can simply be run in the browser, that's why initially it was not encoded.

PR #500, opened 2022-10-07
Document FRONT_COMMERCE_CART_CACHE_ENABLE and cart cache feature

dpobel:
https://deploy-preview-500--heuristic-almeida-1a1f35.netlify.app/docs/magento2/advanced#caching-customers-cart
https://deploy-preview-500--heuristic-almeida-1a1f35.netlify.app/docs/reference/environment-variables#cache


real34 (changes requested):
The current content looks good to me.

I however think that we miss explanations about what to do and verify in _userland code_ when activating this cache feature.

I suggest to add a section documenting `decorateWithCartCacheExpire` and when it must be used. A few sentences might be sufficient. Then linking to https://gitlab.com/front-commerce/front-commerce/-/blob/0869a15affee6173e5ffc3894fc7b64b90ff64d7/src/server/modules/magento2/cart/loaders/CachedCartLoader.js might be good enough to help integrators understand where they must look in their app.

real34 (approved):
Thanks! :ok_hand:

PR #555, opened 2022-12-02
ImageAdapter - TwicPics

paul-vd:
https://deploy-preview-555--heuristic-almeida-1a1f35.netlify.app/docs/advanced/images-adapters/twicpics

### DIFF
--- a/docs/advanced/images-adapters/twicpics.mdx
+++ b/docs/advanced/images-adapters/twicpics.mdx
@@ -1,0 +1,129 @@
+---
+sidebar_position: 1
+title: TwicPics
+description:
+  Deliver perfectly sized images in real-time to create visual-rich experiences.
+---
+
+import ContactLink from "@site/src/components/ContactLink";
+import SinceVersion from "@site/src/components/SinceVersion";
+
+<SinceVersion tag="2.21" />
+
+<p>{frontMatter.description}</p>
+
+## Installation
+
+Add the `twicpics` module to your config file:
+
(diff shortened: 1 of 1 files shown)


docs/advanced/images-adapters/twicpics.mdx
```diff
@@ -0,0 +1,97 @@
...
+description:
+  Deliver perfectly sized images in real-time to create visual-rich experiences.
+tags:
+  - Image Adapter
+  - TwicPics
+  - Since version 2.20.x
+---
+
+import ContactLink from "@site/src/components/ContactLink";
+
```
real34, line 13: similar to Cloudinary's: there is a missing `<SinceVersion>` component here.

If we want to use tags for it (could be a good idea), I think it has to be done separately and globally.
Possibly, we could automate the `<SinceVersion>` at layout level if we detect the tag of a version.

That way, we could have no duplication and consistency.

But this is out of scope of this PR, hence the suggestion of adding the component manually like everywhere else, for consistency.

real34 (approved):

docs/advanced/images-adapters/twicpics.mdx
paul-vd: Yea so this was prior to the `SinceVersion` component which led to the creation of that component.

I still need to merge it to the next branch and rebase, I will handle that next, just finishing up the demo's

real34:
This MR is blocked until the feature is shipped in a release.

PR #572, opened 2023-01-10
feat: add CORS headers to allow RSS inclusion everywhere

real34:
This PR allows our RSS feeds to be accessed by other websites client sides.

Example: https://developers.front-commerce.com/changelog/rss.xml


real34:
I'm merging as I need it right away, but don't hesitate to provide feedbacks.

PR #576, opened 2023-01-18
fix(twicpics): update domain config key

paul-vd:


paul-vd:
I'm not waiting for reviews as this is a fix to a key that was initially used, but changed in the final version to better match the terminologie used by TwicPics.

PR #594, opened 2023-02-21
feat(attraqt-recommendation): Added Attraqt Chrome extension setup and example

Kagaminara:
This MR adds the setup of Attraqt's Orchestrator Chrome extension, and also provides an example of utilisation of the loader  provided by Front-Commerce for this usecase.

### Discussions

As suggested in the comments on this PR, Front-Commerce could also provides with a module that basically does what the examples shows (+ `context`), and this should cover most of the usages of Attraqt Recommendation API.

### DIFF
--- a/docs/magento2/search-engine.mdx
+++ b/docs/magento2/search-engine.mdx
@@ -370,14 +370,14 @@ query Search {
 If you are using the default theme or the Chocolatine theme, the search bar
 should now be visible.
 
-## Attraqt recommendations
+### Attraqt recommendations
 
 > _Since version `2.19`_
 
-Front-Commerce provides a loaders to retrieve recommendations from
+Front-Commerce provides a loader to retrieve recommendations from
 [Attraqt's Widgets](https://attraqt.gitbook.io/developer-documentation/xo-recommendations/readme-first).
 
-### Front-Commerce configuration
+#### Front-Commerce configuration
 
 To enable the Attraqt recommendation module, you need to add it to your
 `.front-commerce.js` config:
@@ -396,6 +396,15 @@ module.exports = {
     // highlight-end
     { name: "Magento2", path: "server/modules/magento2" },
   ],
+  webModules: [
+    { name: "FrontCommerce", path: "./src/web" },
+    // highlight-start
+    {
+      name: "AttraqtRecommendation",
+      path: "front-commerce/modules/recommendations-attraqt/web",
+    },
+    // highlight-end
+  ],
 };
 ```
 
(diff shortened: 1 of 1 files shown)


docs/magento2/search-engine.mdx
```diff
@@ -370,14 +370,14 @@ query Search {
 If you are using the default theme or the Chocolatine theme, the search bar
 should now be visible.
 
-## Attraqt recommendations
+### Attraqt recommendations
```
Kagaminara, line 373: Relocated Attraqt Recommendations inside the `Attraqt` block

docs/magento2/search-engine.mdx
```diff
@@ -405,8 +414,94 @@ Then, you need to define the API url in your environment variables:
...
+
+:::
+
+You will first need to update your GraphQL schema to implement this interface
+and use it in a query:
+
+```graphql title="my-module/server/modules/SuggestedProducts/schema.gql"
+type AttraqtRecommendation implements AttraqtRecommendationResult {
+  id: ID
+}
```
Kagaminara, line 438: Actually product recommendation should be retrieved here as well, like so:
```suggestion
type AttraqtRecommendation implements AttraqtRecommendationResult {
  id: ID
  recommendation: [AttraqtRecommendation]
}
```

docs/magento2/search-engine.mdx
```diff
@@ -405,8 +414,94 @@ Then, you need to define the API url in your environment variables:
...
+
+extend type Query {
+  attraqtRecommendation(attraqtWidgetId: String!): AttraqtRecommendation
```
Kagaminara, line 441: For the sake of the exemple, this query only uses the widget ID, however in a more real usecase the `context` should be passed as well.

docs/magento2/search-engine.mdx
```diff
@@ -405,8 +414,94 @@ Then, you need to define the API url in your environment variables:
 FRONT_COMMERCE_ATTRAQT_RECOMMENDATION_API_URL=http://api.early-birds.io/widget
 ```
 
-After this step, you should be able to use this loader in your modules to fetch
-Attraqt recommendations for your widgets.
+#### Using Attraqt Orchestrator Chrome extension
+
+In order to use this Attraqt Orchestrator extension for Google Chrome,
```
real34, line 419:
```suggestion
In order to use the [Attraqt Orchestrator extension for Google Chrome](https://chrome.google.com/webstore/detail/attraqt-experience-orches/icognleckonecpphokboemgcgcohcepi),
```

docs/magento2/search-engine.mdx
real34: It should even be a `ProductRecommendation` I think. Not mentioning Attraqt except in the implemented interface would make it a bit cleaner.

docs/magento2/search-engine.mdx
real34: I'd suggest to even remove the `attraqtWidgetId` and consider it has a hardcoded value called in the resolver (with context if needed), as an example.

```suggestion
  productRecommendation: ProductRecommendation
```

real34 (changes requested):

real34 (changes requested):

docs/magento2/search-engine.mdx
```diff
@@ -405,8 +414,103 @@ Then, you need to define the API url in your environment variables:
...
+          "62f3c9d50cc94f235967672f"
+        );
+      const recommendedProductSku = recommendation.recommendations?.[0]?.id;
+      const product = loaders.Product.load(recommendedProductSku);
+      return {
+        ...recommendation,
+        product,
+      };
+    },
+  },
```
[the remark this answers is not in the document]
Kagaminara: That's where it becomes weird. In the real world, `id` might not actually be the SKU, all the objects inside `recommendations` entirely depends on what is configured in Attraqt (by Attraqt), so I hesitated a bit before putting it as an example in our docs.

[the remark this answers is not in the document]
real34: Yes, I understand. It might be clearer to indeed use another field name to make this clearer.

Maybe something like this:
```
// Note: the shape of recommendation.recommendations will vary
// depending on your data indexed in Attraqt. Adapt this example to your use case.
const skus = recommendation.recommendations.map(({ sku }) => sku);
```

real34 (approved):
Thanks for the fixups and documentation! Let's ship it :rocket: :smile:

PR #599, opened 2023-03-01
feat(rma): document known issues

paul-vd:
https://deploy-preview-599--heuristic-almeida-1a1f35.netlify.app/docs/magento2/commerce#rma


docs/magento2/commerce.mdx
```diff
@@ -185,3 +212,180 @@ filters on ` option_id` instead of `value_id`.
...
+```mdx-code-block
+  </div>
+</details>
+```
+
+<!--------------------DETAILS-------------------->
+
+```mdx-code-block
+<details>
+  <summary className="flex justify-between"><span>The <code>Yes/No</code> field returns the id instead of value to the end user</span><SinceVersion inline tag="2.22.1" /></summary>
```
dpobel, line 225: I find a bit odd to specify a FC version in a known issue so I would suggest to just remove it from here and leave the version in paragraph stating the field types support.

```suggestion
  <summary className="flex justify-between"><span>The <code>Yes/No</code> field returns the id instead of value to the end user</span></summary>
```

docs/magento2/commerce.mdx
```diff
@@ -185,3 +212,180 @@ filters on ` option_id` instead of `value_id`.
...
+```mdx-code-block
+  </div>
+</details>
+```
+
+<!--------------------DETAILS-------------------->
+
+```mdx-code-block
+<details>
+  <summary className="flex justify-between"><span>The <code>File and Image</code> fields are not supported</span><SinceVersion inline tag="2.22.1" /></summary>
```
dpobel, line 345: same here on the version.

I would even argue that it's not a known issue but rather the explanation of why File and Image are not supported so maybe you could move the sentence _Magento is not yet ready to handle such files in its current state._ to the table on Field type support.

docs/magento2/commerce.mdx
```diff
@@ -185,3 +212,180 @@ filters on ` option_id` instead of `value_id`.
...
+```
+
+#### Returns Attributes
+
+You can learn more about configuring Returns Attributes on the
+[Magento Commerce documentation](https://experienceleague.adobe.com/docs/commerce-admin/stores-sales/order-management/returns/attributes-returns.html)
+
+##### Supported Input Types
+
+<SinceVersion tag="2.22.1" />
```
dpobel, line 382: that should be specified for each type because the Dropdown was supported before
that version.

Also, unless I'm mistaken this won't be backported to 2.22 so the version should
be 2.23

dpobel (commented):
while on this page, you can maybe also fix the line break at the end of the introduction

image

PR #608, opened 2023-03-10
feat: add blog comments with giscus

paul-vd:
### Why?
This allows us to communicate with the readers on comments/questions they might have in the blog post.

### What?
It adds this comments section at the bottom which is driven through the GitHub Discissions api

### Example
https://deploy-preview-608--heuristic-almeida-1a1f35.netlify.app/blog/wsl-development-environment#conclusion

image


real34 (approved):
Codewise it looks good.

I'm not specifically sure it will bring value though, as discussions might more happen elsewhere however we can give it a try if everyone agrees.

We will reevaluate after some time.

PR #619, opened 2023-04-04
Links to source code

dpobel:
Summary:

* when the link targets a specific version (changelog, migration guide, …) the link is updated to lead to the file in the `2.n.x` branch
* when the link targets a directory, the link is updated to the `2.x` branch
* in other case, the link is the permalink to the file in the current `2.x` branch ie the commit that configure the CI for `2.x`

along the way, I fixed some hash to target the right lines and removed the link IntlDecorator that was removed like 3 years ago.


real34 (approved):
:pray: Thanks a lot for tackling this!

Good idea to use `2.y.x` links for some of them!

PR #623, opened 2023-04-05
chore: add modal route manual migration

paul-vd:
https://deploy-preview-623--heuristic-almeida-1a1f35.netlify.app/docs/remixed/migration/manual-migration#modal-routes

### DIFF
--- a/docs/migration/manual-migration.mdx
+++ b/docs/migration/manual-migration.mdx
@@ -81,3 +81,35 @@ You would need to import the `color.scss` directly in the `Example.scss` file
   color: darken($colorPrimary, 10%);
 }
 ```
+
+## Modal Routes
+
+In V2, a modal state parameter facilitated rendering arbitrary routes in a modal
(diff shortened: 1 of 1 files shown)


docs/migration/manual-migration.mdx
```diff
@@ -81,3 +81,35 @@ You would need to import the `color.scss` directly in the `Example.scss` file
   color: darken($colorPrimary, 10%);
 }
 ```
+
+## Modal Routes
+
+In V2, a modal state parameter facilitated rendering arbitrary routes in a modal
+during routing. In Remix, this is non-trivial or impossible unless the target
+route is a child route. We have introduced a ModalLink component to replace the
```
real34, line 89:
```suggestion
during routing. In Remix, this is non-trivial unless the target
route is a child route. We have introduced a `ModalLink` component to replace the
```

real34 (approved):

PR #641, opened 2023-04-20
chore: document legacy wysiwyg removal

paul-vd:
https://deploy-preview-641--heuristic-almeida-1a1f35.netlify.app/docs/remixed/migration/deprecated-code-removal#wysiwygv1

### DIFF
--- a/docs/migration/deprecated-code-removal.mdx
+++ b/docs/migration/deprecated-code-removal.mdx
@@ -47,6 +47,47 @@ The legacy analytics implementation based on analytics.js has been removed in
 [Remove legacy
 analytics](https://gitlab.blackswift.cloud/front-commerce/front-commerce/-/merge_requests/1864).
 
+### WysiwygV1
+
+The legacy `WysiwygV1` implementation has been removed in
+[Remove legacy WysiwygV1](https://gitlab.blackswift.cloud/front-commerce/front-commerce/-/merge_requests/1929)
+
+<!--TODO: add link to MR which applies updates the location of WysiwygV2 -> Wysiwyg and applies the codemod changes-->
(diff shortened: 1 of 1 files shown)


docs/migration/deprecated-code-removal.mdx
```diff
@@ -47,6 +47,39 @@ The legacy analytics implementation based on analytics.js has been removed in
...
 analytics](https://gitlab.blackswift.cloud/front-commerce/front-commerce/-/merge_requests/1864).
 
+### WysiwygV1
+
+The legacy `WysiwygV1` implementation has been removed in
+[Remove legacy WysiwygV1](https://gitlab.blackswift.cloud/front-commerce/front-commerce/-/merge_requests/1929)
+
+<!--TODO: add link to MR which applies updates the location of WysiwygV2 -> Wysiwyg and applies the codemod changes-->
+
+All related imports will be automatically updated after running the [automated migration process](/docs/remixed/migration/automated-migration) step. for example:
```
real34, line 57: This is not clear to me (and thus I think for integrators).

What is really needed is:
1. a manual migration step (from WYSIWYG v1 to WYSIWYG v2) as a prerequisite
2. running the codemod

The codemod could die if it detects an incorrect import to WYSIWYG, but it must be stated more clearly that the project must not use WYSIWYG v1 in a first place (it could be as simple to check as searching for the import string)

real34 (changes requested):

docs/migration/deprecated-code-removal.mdx
paul-vd: Updated the text, please re-review

real34 (approved):
It's better indeed!

Looks good to me. We could refine if it isn't clear for integrators from feedback.

PR #649, opened 2023-05-02
fix: properly document how to enable Payment on account for a company user

dpobel:
Being able to use the Payment on Account payment method requires some
permissions within the company. This PR adds a description of those (haven't
found the corresponding documentation in Magento's doc)

### DIFF
--- a/versioned_docs/version-2.x/advanced/payments/payment-on-account.mdx
+++ b/versioned_docs/version-2.x/advanced/payments/payment-on-account.mdx
@@ -28,6 +28,17 @@ This feature is directly provided by Adobe Commerce. Please refer to
 
 :::
 
+In addition, this feature is also controlled by the roles and the associated
+permissions assigned to a user within the company. For now, those roles can only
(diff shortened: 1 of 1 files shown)


versioned_docs/version-2.x/advanced/payments/payment-on-account.mdx
```diff
@@ -28,6 +28,17 @@ This feature is directly provided by Adobe Commerce. Please refer to
 
 :::
 
+In addition, this feature is also controlled by the roles and the associated
+permissions assigned to a user within the company. For now, those roles can only
+be set through the Magento frontend. So to make sure a company user can use the
```
real34, line 33: **issue:** It isn't viable to suggest to proceed like this for production use cases.

**suggestion:** mention programatical  roles management for production, and highlight that the procedure is for local developments only (or dev)

```suggestion
be set programmtically or through the Magento frontend. In local environments, to make sure a company user can use the
```

real34 (approved):

PR #663, opened 2023-05-22
Added a tip on the usage of https so that Google Pay and Apple Pay show up

dpobel:


versioned_docs/version-2.x/advanced/payments/stripe.mdx
```diff
@@ -93,6 +93,13 @@ Google Pay with Stripe](https://stripe.com/docs/google-pay) and [the Apple Pay
 with Stripe](https://stripe.com/docs/apple-pay) documentation to learn how to
 configure those payment methods.
 
+:::tip
+
+For those payment methods to show up, your Front-Commerce store must use the
+https protocol even in a test environment.
```
real34, line 99: **question:** is it also needed when testing under http://localhost:PORT?

I'm not sure, but I think I could use Google Pay locally under http.

dpobel: yes it's needed (at least with Stripe), this is exactly the issue I had yesterday that led to this MR :)

PR #669, opened 2023-05-25
Update hipay.mdx

paul-vd:
## Why?
When reading the docs, it's not clear that the custom loader is optional, I rephrased it to be more clear on that detail.

I also changed the `warning` citation to `caution`.


real34:
_I merged because it only seem that we forgot about this one, and wasn't worth bothering @paul-vd to click on a button :smile: ._

PR #686, opened 2023-06-27
Better document bgColors / bg image parameter

dpobel:
the URL to the image is wrong, the query parameter is `bg` not `bgColors` also I try to add a full example with `bg` so that it's easier to understand.


Kagaminara (approved):

real34 (approved):
:tada: Thanks for looking into this.

PR #696, opened 2023-07-04
Documented canonical url computation changes

Kagaminara:
See [related MR](https://gitlab.blackswift.cloud/front-commerce/front-commerce/-/merge_requests/2121).

### DIFF
--- a/versioned_docs/version-2.x/appendices/migration-guides/index.mdx
+++ b/versioned_docs/version-2.x/appendices/migration-guides/index.mdx
@@ -14,6 +14,24 @@ import TabItem from "@theme/TabItem";
 
 ## `2.24.0` -> `2.25.0`
 
+### Updated canonical URL creation method
+
+**Breaking**: In this version we have changed the way
+`QueryHelper.filterCanonicalQueryStringParameters` compute the canonical URL
+parameters, in order to better fit SEO-recommended rules (see
+[Google article](https://developers.google.com/search/docs/crawling-indexing/canonicalization#:~:text=the%20results%20of%20sorting%20and%20filtering%20functions%20of%20a%20category%20page)
+on this matter).
+
+To opt-in to this behaviour, you can pass a boolean as the second parameter of
+`QueryHelper.filterCanonicalQueryStringParameters` to configure whether or not
+additional filters should be included in the canonical. `true` (default value)
+will include additional filters, and `false` will exclude them. Please note that
+we removed the `sorting`, `minPrice` and `maxPrice` parameters from the
+canonical url parameters.
+
+See more in the
+[related commit](https://gitlab.blackswift.cloud/front-commerce/front-commerce/-/commit/ebffdc8412f2657bff6782bab1c49568fc78520a?merge_request_iid=2121).
+
 ### Drop support of Node 14
 
 In this version we've **dropped support for Node 14**


versioned_docs/version-2.x/appendices/migration-guides/index.mdx
```diff
@@ -14,6 +14,22 @@ import TabItem from "@theme/TabItem";
...
+**Breaking**: In this version we have changed the way
+`QueryHelper.filterCanonicalQueryStringParameters` compute the canonical URL
+parameters, in order to better fit SEO-recommended rules (see
+[Google article](https://developers.google.com/search/docs/crawling-indexing/canonicalization#:~:text=the%20results%20of%20sorting%20and%20filtering%20functions%20of%20a%20category%20page)
+on this matter).
+
+We removed sorting from the URL parameters completely, and added a new parameter
+(`shouldQueryIncludeAdditionalParameters`) to configure whether or not
+additional filters should be included in the canonical query string parameters.
+This new parameter is defaulted to `true` for backward compatibility.
```
paul-vd, line 28: Based on my comment here: https://gitlab.blackswift.cloud/front-commerce/front-commerce/-/merge_requests/2121#note_87169, I think this should be rephrased, because the wording makes it seems like `shouldQueryIncludeAdditionalParameters=true` is equivalent to an opt-out, but it's not, because there are other params.


Replace `XXXXX` with the action required for integrators (see [suggestion](https://gitlab.blackswift.cloud/front-commerce/front-commerce/-/merge_requests/2121#note_87178))

```suggestion
To opt-out of this behaviour, you can XXXXX  to configure whether or not
additional filters should be included in the canonical, please note that we removed the `sorting`, `minPrice` and `maxPrice` parameters from the canonical
```

paul-vd (changes requested):

versioned_docs/version-2.x/appendices/migration-guides/index.mdx
Kagaminara: Wouldn't it be `to opt-in` instead ? Since by default the behavior did not change from previous versions ?

Kagaminara: Would something like this fit the need ? 
```suggestion
To opt-in this behaviour, you can pass a boolean as the second parameter of
`QueryHelper.filterCanonicalQueryStringParameters` to configure whether or not
additional filters should be included in the canonical. `true` (default value)
will include additional filters, and `false` will exclude them. Please note that
we removed the `sorting`, `minPrice` and `maxPrice` parameters from the
canonical.
```

paul-vd: That's much better

versioned_docs/version-2.x/appendices/migration-guides/index.mdx
```diff
@@ -14,6 +14,24 @@ import TabItem from "@theme/TabItem";
...
 
+### Updated canonical URL creation method
+
+**Breaking**: In this version we have changed the way
+`QueryHelper.filterCanonicalQueryStringParameters` compute the canonical URL
+parameters, in order to better fit SEO-recommended rules (see
+[Google article](https://developers.google.com/search/docs/crawling-indexing/canonicalization#:~:text=the%20results%20of%20sorting%20and%20filtering%20functions%20of%20a%20category%20page)
+on this matter).
+
+To opt-in this behaviour, you can pass a boolean as the second parameter of
```
paul-vd, line 25:
```suggestion
To opt-in to this behaviour, you can pass a boolean as the second parameter of
```

paul-vd (approved):

PR #713, opened 2023-08-01
Document the `npm install` command to upgrade FC in a project

real34:
This PR brings a minor improvement to the Migration guide: it adds a reminder with the command integrators could use to update FC in their project.

It is a command that we're regularly asked about, and it felt to be a good idea to have it here.

**[Preview](https://deploy-preview-713--heuristic-almeida-1a1f35.netlify.app/docs/2.x/appendices/migration-guides/)**


real34:
Merging as I'm quite confident with this change, and to prevent additional delays or interruptions for teammates. I'll follow any feedbacks on this PR and could issue followup PRs if needed.

PR #722, opened 2023-08-07
Document inlining of `getPriceWithSelectedGroupedItems` and `fusePriceWithBundleOptions`

Kagaminara:
This MR document a manual change we made since those two method were only used in `packages/magento2/modules/catalog/products/resolvers.js`.


real34:
I'm awaiting a bit before reviewing and approving this, because I want to reevaluate the reasons behind it and see if that makes sense to mention it. Indeed, resolvers are our own internal API (so the refactoring shouldn't impact integrators).

Kagaminara:
Closing as it only concerns our internal API.

PR #730, opened 2023-08-17
Document Link component manual migration

Kagaminara:
Adds documentation for checking \<Link> components usage during manual migration process.

Related MR comment: https://gitlab.blackswift.cloud/front-commerce/front-commerce/-/merge_requests/2291#note_94304

### DIFF
--- a/docs/migration/manual-migration.mdx
+++ b/docs/migration/manual-migration.mdx
@@ -21,9 +21,8 @@ The `xxxSeo` components (eg: `HomeSeo`) have been removed in favor of
 If you use SEO components or have overriden Seo components, you need to:
 
 - Remove import and usage of the component from your page
-- Use the [meta](https://remix.run/docs/en/main/route/meta-v2)
-  function with our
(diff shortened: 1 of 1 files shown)


docs/migration/manual-migration.mdx
```diff
@@ -311,14 +311,30 @@ following stylesheet:
...
 // add-next-line
 @import "theme/some/path/defining/another-selector";
 
 .example {
   @extend .another-selector;
 }
-```
+## Link component
+
+In V3, we now use Remix's own `Link` component from `@remix-run/react`. Although
```
real34, line 323: **issue:** we do, but developers shouldn't. They must keep using our atom. Can you detail the reasoning behind it? Any examples?

Kagaminara: Indeed, I actually wrote a doc on something absolutely unrelated to the real hightlighted issue in the comment... My brain was elsewhere it seems, I'll fix it.

Kagaminara: After investigation the related change mentioned in the description was not needed at all, and so does this MR. I'll close it.

PR #762, opened 2023-09-22
Bad loggingService import path

bastienlm:
Try on version 2.20, but the path for import loggingService is invalid, we need to use `server/core/logs/loggingService`


paul-vd (approved):

paul-vd:
Nice catch, thanks @bastienlm for your contribution! 🥳

PR #768, opened 2023-09-28
New documentation layout / structure

Kagaminara:
This MR aims to open discussions about layout choices for the new V3 documentation. Once validated and merged, it then can be enriched (and completed) by smaller incrementations.

It provides and "introduction" that explains how to do a basic install of Front-Commerce, as well as 6 "main" categories, which, for now, only contains studs of what will be featured in them:
- `Get started`: this section will provide tutorials on how to do things with Front-Commerce. This is the equivalent of testing things on a "Naked FC".
- `Guides`: this section will provide guides on how do to "_working_" / "_usable_" tasks in Front-Commerce.
- `Extensions`: This is the list of non-core extensions provided by Front-Commerce. Each listed extension will have (most likely) two "folders":
  - `How-to` which will provide tutorials on how to gets tasks done that are specific to the related extension
  - `Reference` which will provide technical details on how the extension's feature are working, and how they can be used.
- `API Reference` will contain all core-related packages, and document how to use the feature they provide
- `Concepts` are a list of mindsets / philosophy concepts we implement at Front-Commerce
- `Migrating from v2`: _(open to debate)_ could also be under a `Migrating` section, which would contain all subjects linked to migrations (i.e.: v2 -> v3, v2.28 ->v2.29, etc.)  

image


docs/03-extensions/_category_.yml
```diff
@@ -0,0 +1,2 @@
+label: Extensions
+collapsible: true
```
mGGk-fr, line 2: What about splitting extensions by type ? Like :
* Extensions
  * Theming
  * Connectors
  * Payment

docs/01-get-started/editor-setup.mdx
```diff
@@ -0,0 +1,17 @@
...
+
+<p>{frontMatter.description}</p>
+
+:::info WIP
+
+Add instructions here:
+
+- VSCode
+- WebStorm
+- Vim
```
dpobel, line 15: :muscle:

docs/03-extensions/_category_.yml
paul-vd: I think that would be too much nesting again, although it technically is correct.

Kagaminara: The inital purpose was to keep documentation of each extension (in the monorepos sense) separated, and not to mix everything like we did in v2's documentation.

dpobel (approved):

docs/03-extensions/_category_.yml
mGGk-fr: LGTM Then :slightly_smiling_face:

paul-vd (approved):

PR #775, opened 2023-09-29
Theme Chocolatine: Document css variables

Kagaminara:
This MR aims to list all variables currently used in the theme Chocolatine and provide a quick brief on what they are and should be used for.

I'm wide open to suggestion about phrasing, as design is not my _forte_. Also, I'm not sure how deep in the variables usage it would be interesting / useful to go.

Note: A lot more variables than what is present in this list are _exported_ in the theme. However, only the variables _used_ in the theme are documented.

---

**[Preview](https://deploy-preview-775--heuristic-almeida-1a1f35.netlify.app/docs/remixed/extensions/theme-chocolatine/reference/css-variables)**

### DIFF
--- a/docs/03-extensions/theme-chocolatine/reference/css-variables.mdx
+++ b/docs/03-extensions/theme-chocolatine/reference/css-variables.mdx
@@ -1,15 +1,99 @@
 ---
 title: "CSS variables"
 description:
-  "This reference lists all the CSS variables useful for customizing the theme
-  Chocolatine."
+  "This reference lists all the CSS variables useful for customizing the theme."
 sidebar_position: 1
 ---
 
 <p>{frontMatter.description}</p>
 
-:::info WIP
+Front-Commerce's theme Chocolatine uses a set of CSS variables defined in
+`theme/_variables.scss` to ensure a certain consistency in the application's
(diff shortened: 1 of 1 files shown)


mGGk-fr (approved):

docs/03-extensions/theme-chocolatine/reference/css-variables.mdx
```diff
@@ -1,15 +1,98 @@
...
+Defined in `theme/components/atoms/Colors/_colors.scss`, color variables are
+useful to ensure a consistent design in your app.
+
+## Base colors variables
+
+Those variables are often used to define other meaningful variables, and serve
+as a root color definition base for your theme. They are used in a wide variety
+of situation in Theme Chocolatine, from text color to backgrounds.
+
+- `white`, `black` and `red`
```
real34, line 28: **nitpick:** I think it'd be clearer if CSS variables had the `--` prefix. It'll be closer to what developers view in their code.

**suggestion:** 
```suggestion
- `--white`, `--black` and `--red`
```

real34 (approved):
Besides the minor feedbacks, it's good enough for a first version. We'll iterate if needed.

Thanks for gathering those.

PR #781, opened 2023-10-02
doc: makeCacheFromRequest

mGGk-fr:
# What

This MR adds documentation for makeCacheFromRequest

# Preview

https://deploy-preview-781--heuristic-almeida-1a1f35.netlify.app/docs/remixed/api-reference/front-commerce-core/make-cache-from-request

### DIFF
--- a/docs/04-api-reference/front-commerce-core/make-cache-from-request.mdx
+++ b/docs/04-api-reference/front-commerce-core/make-cache-from-request.mdx
@@ -1,0 +1,64 @@
+---
+title: "makeCacheFromRequest"
+sidebar_position: 5
+description: "The makeCacheFromRequest allows to cache response for a request."
+---
+
+<p>{frontMatter.description}</p>
+
+## Usage
+
+```js
+makeCacheFromRequest(request, cacheImplementation);
+```
+
+Arguments:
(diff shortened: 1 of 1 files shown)


docs/04-api-reference/front-commerce-core/make-cache-from-request.mdx
```diff
@@ -0,0 +1,64 @@
+---
+title: "makeCacheFromRequest"
+sidebar_position: 5
+description: "The makeCacheFromRequest allows to cache response for a request."
```
real34, line 4: **issue:** this is not correct at all. It's a factory for the caching layer.

Several locations of the documentation below is also incorrect.

**suggestion:** @mGGk-fr do you want to proof-read it, or get detailed feebdack on it (in this case it'll likely be faster for me to do another MR fixing this)

real34 (changes requested):

docs/04-api-reference/front-commerce-core/make-cache-from-request.mdx
mGGk-fr: Added task FC-1754 for follow up

PR #786, opened 2023-10-04
fix: broken link on CSS variables reference

real34:
Regression introduced in #775.
The build currently is broken on `main`:  https://app.netlify.com/sites/heuristic-almeida-1a1f35/deploys/651d55e1586ef40008be0d2e

This patch links to "Customize the styles" page instead. I guess that was the initial intent.


Kagaminara (approved):
Thanks !

PR #795, opened 2023-10-10
chore: document the useApiFetcher

paul-vd:
## What?
This documents the `useApiFetcher` introduced in https://gitlab.blackswift.cloud/front-commerce/front-commerce/-/merge_requests/2691

## Preview
- https://deploy-preview-795--heuristic-almeida-1a1f35.netlify.app/docs/remixed/api-reference/front-commerce-remix/#useapifetcher


docs/04-api-reference/front-commerce-remix/index.mdx
```diff
@@ -31,25 +33,56 @@ the `@front-commerce/remix` reference to your `remix.env.d.ts` file.
 /// <reference types="@front-commerce/remix" />
 ```
 
-### Improved loader data types
```
paul-vd, line 34: I removed this as it resulted in reduced visibility for the `@front-commerce/remix/node` and  `@front-commerce/remix/react` api's this also now allows for subtitles in each of those sections.

docs/04-api-reference/front-commerce-remix/index.mdx
```diff
@@ -31,25 +33,56 @@ the `@front-commerce/remix` reference to your `remix.env.d.ts` file.
...
+  })
+);
+```
 
-#### `@front-commerce/remix/node`
+## `@front-commerce/remix/node`
 
 - `json`
 - `redirect`
 - `defer`
```
paul-vd, line 71: I should probably add links to remix for these as well?

real34: Yes, that'd be great

docs/04-api-reference/front-commerce-remix/index.mdx
```diff
@@ -91,33 +124,75 @@ Instead of the standard remix `SerializeObject` type:
...
+    type="video/mp4"
+  />
+  Your browser does not support the video tag.
+</video>
+
+</BrowserWindow>
+
+As you can see with the `useFetcher` it will call the api for each component,
+with the `useApiFetcher` it will only called the api once and share the data
+between the components.
```
real34, line 198:
```suggestion
As you can see with the `useFetcher` it will call the API for each component,
with the `useApiFetcher` it will only called the API once and share the data
between the components.
```

real34: Also, don't you think it'd be worth to add a line about revalidation? (it will also prevent multiple calls in this case)

docs/04-api-reference/front-commerce-remix/index.mdx
```diff
@@ -91,33 +124,75 @@ Instead of the standard remix `SerializeObject` type:
...
+  return (
+    <div>
+      Cart Id: {data.cart.id}
+      Cart Qty: {data.cart.items_qty}
+    </div>
+  );
+}
 ```
+
+Here is an example of 5 concurrent calls to the same `/api/cart` endpoint:
```
real34, line 176: **nitpick:** since it uses the same component than above, it could be great to provide a working example of this

**suggestion:** add the code for the "button displaying the cart components" in a `<details>` here

**suggestion(alt):** create a working example for this and link from here. This could be postponed, but if you still have the code around it could be dumped in the Linear issue tracking this.

real34 (approved):

docs/04-api-reference/front-commerce-remix/index.mdx
```diff
@@ -91,33 +124,75 @@ Instead of the standard remix `SerializeObject` type:
...
-## Front-Commerce Context
+### `useApiFetcher`
 
-To ensure the Front-Commerce Context is available in your loaders and actions,
-you need to add the `createFrontCommerceContext` to your server via the
-[`getLoadContext`](https://remix.run/docs/en/2.0.1/route/loader#context), to do
-this, update your
-[`server.ts`](https://gitlab.blackswift.cloud/front-commerce/front-commerce/-/blob/b99f0475ce494591a54fdc2b904e956a083e9b51/skeleton/server.ts)
+The `useApiFetcher` is an optimised fetcher which allows to deduplicate client
+side calls the the same `/api` endpoint. You can replace a load of `useFetcher`
```
Kagaminara, line 130: > You can replace a load of `useFetcher` calls

I'm not sure what this means, did you meant "a lot of `useFetcher` calls"?

docs/04-api-reference/front-commerce-remix/index.mdx
```diff
@@ -91,33 +124,75 @@ Instead of the standard remix `SerializeObject` type:
...
+  const data = fetcher.data
+
+  return (
+    <div>
+      Cart Id: {data.cart.id}
+      Cart Qty: {data.cart.items_qty}
+    </div>
+  );
+}
 ```
```
Kagaminara, line 174: **nit**: I find the diff harder to read than a good old "before / after" block. Plus, I think this would be a bit easier to see the difference in terms code size.

Kagaminara (approved):

docs/04-api-reference/front-commerce-remix/index.mdx
paul-vd: I will rephrase, it, I make reference to the `load` method returned from the `useFetcher` hook, but after re-reading I notice it's not clear.

Kagaminara: Ooh makes sense. Thanks !

PR #799, opened 2023-10-17
chore: document makeUserClientFromRequest and makeCartUrlBuilderFromRequest

paul-vd:
## What?
Documents the axois apis for magento2, see https://gitlab.blackswift.cloud/front-commerce/front-commerce/-/merge_requests/2734

### DIFF
--- a/docs/03-extensions/magento2/reference/axios-instances.mdx
+++ b/docs/03-extensions/magento2/reference/axios-instances.mdx
@@ -10,7 +10,66 @@ sidebar_position: 1
 
 :::info WIP
 
-- `makeUserClientFromRequest`
 - `makeAdminClientFromRequest`
 
 :::info
+
+## `makeUserClientFromRequest`
+
+The `makeUserClientFromRequest` function creates an `axios` instance that can be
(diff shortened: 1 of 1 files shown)


docs/03-extensions/magento2/reference/axios-instances.mdx
```diff
@@ -10,7 +10,66 @@ sidebar_position: 1
 
 :::info WIP
 
-- `makeUserClientFromRequest`
 - `makeAdminClientFromRequest`
 
 :::info
+
+## `makeUserClientFromRequest`
```
Kagaminara, line 17: **thought:** I'm not sure to understand what the "user client" would be. It this to be differentiated with an "admin client"?

Kagaminara (approved):
Other than the question, looks good to me.

docs/03-extensions/magento2/reference/axios-instances.mdx
paul-vd: It's to create a request for customers, for example fetching the cart does not require admin access

PR #830, opened 2024-01-15
chore: add release schedule

paul-vd:
## What
Adds our release schedule to the release notes

## Preview
- https://deploy-preview-830--heuristic-almeida-1a1f35.netlify.app/docs/3.x/upgrade/release-notes#release-schedule
- https://deploy-preview-830--heuristic-almeida-1a1f35.netlify.app/docs/2.x/appendices/release-notes#release-schedule


versioned_docs/version-2.x/appendices/release-process.mdx
```diff
@@ -37,15 +37,30 @@ seamless as possible.
 > Additional labels for pre-release can (and in our case will) be added to the
 > MAJOR.MINOR.PATCH format.
 
+## Version Support Policy
+
+Front-Commerce adheres to a dynamic version support policy designed to balance
+innovation with reliability. Specifically, "Front-Commerce supports the last 8
+minor versions at any given point in time." This means at any moment, the
```
Kagaminara, line 44: **thought:** I always find it neat to precise that "8" of our minor versions actually means 1 year of support for each minor version, not sure if you want to include that or not. Non-blocking in any way.

Kagaminara (approved):

versioned_docs/version-2.x/appendices/release-process.mdx
real34: I agree.

What would be more expressive for readers (IMO) would be a table of each version and their support policy (release date -> end of support date -> end of life).

While the end of support is 1 year, it's only for "systematic patch releases". We do support a version for 2 years contractually, in the sense that if one reports a bug we'll fix it.

PR #850, opened 2024-02-26
feat(theme): add scrollToTop automatic documentation

mGGk-fr:
# What

This explain how to disable auto scroll to top when navigating


Kagaminara (approved):
LGTM.

A thought however, should we also explain how to manually enable the scroll to top on specific pages if needed?

paul-vd (approved):

PR #859, opened 2024-03-15
Fc 2044 vite refactor graphql modules api reference

mGGk-fr:
# What

This MR adds reference for two public API:

- `createGraphQLModule` [Preview](https://deploy-preview-859--heuristic-almeida-1a1f35.netlify.app/docs/3.x/api-reference/front-commerce-core/createGraphQLModule)
- `createGraphQLRuntime` [Preview](https://deploy-preview-859--heuristic-almeida-1a1f35.netlify.app/docs/3.x/api-reference/front-commerce-core/createGraphQLRuntime)


paul-vd (approved):
Looks good, I think we also need a `How To` documentation to explain things like
- Adding in the extensions definition
- If typeDefs are included do not add that `graphql::schema` in the extension definition

mGGk-fr:
> Looks good, I think we also need a `How To` documentation to explain things like
> 
>     * Adding in the extensions definition
> 
>     * If typeDefs are included do not add that `graphql::schema` in the extension definition

Thanks for your feedback!
It's part of the issue and this will be handled in another PR :)

paul-vd (changes requested):
**nitpick:** I don't think this is the right place for this, we already have the page[ `graphql` ](https://deploy-preview-859--heuristic-almeida-1a1f35.netlify.app/docs/3.x/api-reference/front-commerce-core/graphql#creategraphqlmodule)under api reference which needs to be updated with this documentation

mGGk-fr:
> **nitpick:** I don't think this is the right place for this, we already have the page[ `graphql` ](https://deploy-preview-859--heuristic-almeida-1a1f35.netlify.app/docs/3.x/api-reference/front-commerce-core/graphql#creategraphqlmodule)under api reference which needs to be updated with this documentation
> 
> Also update API's which no longer exist, [see reference](https://gitlab.blackswift.cloud/front-commerce/front-commerce/-/blob/ed8f406c5ab2101330005cdbd7a7a9233de7c77e/packages/core/graphql/index.ts) for example:
> 
>     * readGraphQLModules - should be removed
> 
>     * schema - should be removed

Reworked with your suggestions

docs/04-api-reference/front-commerce-core/graphql.mdx
```diff
@@ -10,43 +10,80 @@ description:
...
 ```ts
-import { schema } from "@front-commerce/core/graphql";
-
-const acmeSchema = await schema();
+import createFooContextEnhancer from "./createFooContextEnhancer";
+import resolvers from "./resolvers";
+
+export default createGraphQLRuntime({
+  resolvers: resolvers,
```
paul-vd, line 80: We are aiming at inlining resolvers for the future, it would be good to lead by example, so instead of importing here we could add a simple resolver which uses Foo?

paul-vd (approved):

docs/04-api-reference/front-commerce-core/graphql.mdx
mGGk-fr: Added inline resolver and a inline typeDefs, let me know your what you think

docs/04-api-reference/front-commerce-core/graphql.mdx
```diff
@@ -10,43 +10,87 @@ description:
...
+import createFooContextEnhancer from "./createFooContextEnhancer";
+import resolvers from "./resolvers";
+
+export default createGraphQLRuntime({
+  resolvers: {
+    Mutation: {
+      getFoo: () => ({
+        id: 1,
+        bar: 2,
+      }),
```
paul-vd, line 85: You can use `loaders.Foo` to simulate the usage of the contextEnhancer

paul-vd (approved):
one nitpick, otherwise looks good

docs/04-api-reference/front-commerce-core/graphql.mdx
mGGk-fr: @paul-vd just to be sure that i've understood what you mean, can you check : https://github.com/front-commerce/developers.front-commerce.com/pull/859/files#diff-dece421bf56e1ac58f9fb0377cca2759ec3ceed9279fc72964edc52c6d2bc491R82

paul-vd: corret

paul-vd (approved):

PR #861, opened 2024-03-18
docs: Updated defineRemixExtension documentation along with the migration guide

Kagaminara:
This MR updates the docs followings changes regarding routes in `definedRemixExtension` API.

### DIFF
--- a/docs/100-upgrade/migration-guides/3.3-3.4.mdx
+++ b/docs/100-upgrade/migration-guides/3.3-3.4.mdx
@@ -173,6 +173,34 @@ export default defineExtension({
 
 :::
 
+### Refactor of `routes`
+
+In this version we updated how `routes` parameter from `defineRemixExtension` is
+resolved. When using folder based routing, you should pass `import.meta.url` to
+the `routes` in your extension definition:
+
+```js
+{
+  name: "acme",
+  theme: "./extensions/acme/theme",
+  routes: import.meta.url,
+}
+```
+
+:::info
+
+`root.tsx` files in your extensions aren't needed anymore, and can now be
+removed.<br /> For more information, see
+[`defineRemixExtension`'s API reference](/docs/3.x/api-reference/front-commerce-remix/defineRemixExtension#routes)
+
+:::
+
+Additionally, we removed the `UNSTABLE_routes` API. If you were previously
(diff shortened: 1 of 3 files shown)


docs/100-upgrade/migration-guides/3.3-3.4.mdx
```diff
@@ -173,6 +173,23 @@ export default defineExtension({
 
 :::
 
+### Refactor of `routes`
+
+In this version we updated how `routes` parameter from `defineRemixExtension`
+was resolve. You can now pass directly `import.meta.url` and the routes from
+your `/routes` directory will automatically be resolved:
```
paul-vd, line 180:
```suggestion
In this version we updated how `routes` parameter from `defineRemixExtension`
is resolved. When using folder based routing, you should pass `import.meta.url` to the `routes` in your extension definition:
```

paul-vd: Also add a note, if they added a placeholder `root.tsx` it can be removed, also check if there is any mention of it in the documentation which needs to be removed

paul-vd (approved):
I don't think I need to review again unless you want me to. Ping me if you have any questions.

docs/100-upgrade/migration-guides/3.3-3.4.mdx
```diff
@@ -173,6 +173,23 @@ export default defineExtension({
...
+{
+  name: "acme",
+  theme: "./extensions/acme/theme",
+  routes: import.meta.url,
+}
+```
+
+For more information, see
+[`defineRemixExtension`'s API reference](/docs/3.x/api-reference/front-commerce-remix/defineRemixExtension#routes)
+
```
real34, line 192: 2 more things must be documented here:
- the fact that extra `root.tsx` in extensions can now be removed
- same for the imports from `UNSTABLE_routes` (and the definitions).

Also, I think that [this section](https://developers.front-commerce.com/docs/3.x/guides/extend-route-layout#re-use-an-existing-layout-route) should be updated too.

real34 (changes requested):
Some missing things, but otherwise it seems correct to me.

real34 (approved):

PR #867, opened 2024-03-25
Sort docs alphabetically.

paul-vd:
## What?
This sorts the docs alphabetically instead of using a manual `sidebar_position`


real34:
The build fails and can't be previewed.

I think it's relevant for some menus, but How-To sections must still be ordered by ourselves so that it follows an order that makes sense.

mGGk-fr (approved):

real34 (approved):
Thanks!

PR #873, opened 2024-04-07
Fix client logger import example for client

JulienPradet:
Hello :wave:

I've noticed this small typo in the 2.x docs.

Take care :hugs:


real34 (approved):
Thanks a lot! :+1:

PR #877, opened 2024-04-12
docs: Update migration guide about missing ESM migration points

Kagaminara:
This MR adds missing migrations point to take into account while migrating to FC 3.4.

Also fixed a typo in the ESM paragraph.

[Preview](https://deploy-preview-877--heuristic-almeida-1a1f35.netlify.app/docs/3.x/migration/manual-migration#transitioning-to-esm)

### DIFF
--- a/docs/migration/manual-migration.mdx
+++ b/docs/migration/manual-migration.mdx
@@ -15,13 +15,15 @@ parts, or debug issues in the migration CLI output.
 
 Starting 3.0, in a effort to modernize the codebase, we transitioned to ESM.
 This migration allows a better alignment with browsers standards, an improved
-static analysis, enhenced language features and overall a better consistency in
+static analysis, enhanced language features and overall a better consistency in
 the JavaScript ecosystem. To learn more about it, see:
 
 - [What is ESM](https://dev.to/iggredible/what-the-heck-are-cjs-amd-umd-and-esm-ikm)
-- [NodeJS docmentation](https://nodejs.org/api/esm.html)
+- [NodeJS documentation](https://nodejs.org/api/esm.html)
 
 ### Updates in your codebase
+
+#### Updating exports
 
 To cope with this transition, you should check that your code does not use
 `module.exports = ` or `module.exports.someNamedExport =` syntax, and replace it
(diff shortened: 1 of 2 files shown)


docs/migration/manual-migration.mdx
```diff
@@ -38,6 +40,24 @@ grep "module.exports" src/ -r
...
+  // ...
+  "name": "my-project",
+  // highlight-next-line
+  "type": "module",
+  //...
+```
+
+#### Uses of `__dirname`
+
+`__dirname` isn't available anymore in ESM. To cope with this change, you will
```
Kagaminara, line 58: Given how variable a use of `__dirname` can be, I wasn't comfortable about giving replacement suggestions here.

paul-vd: What do you mean by "variable" in esm it just replaced with meta

```ts
import { dirname } from 'path';
import { fileURLToPath } from 'url';

const __dirname = dirname(fileURLToPath(import.meta.url));
```

paul-vd (approved):

docs/migration/manual-migration.mdx
```diff
@@ -38,6 +40,24 @@ grep "module.exports" src/ -r
...
+This transition will also require your project to be declared as a `module`. In
+order to do so, you can add the following to your `package.json`:
+
+```json
+  // ...
+  "name": "my-project",
+  // highlight-next-line
+  "type": "module",
+  //...
+```
```
real34, line 54: **issue:** I don't see the value of such information here.\
When migrating from 2.x, integrators will start with a fresh 3.x skeleton. The skeleton already has it.

**suggestion:** remove this part.

**suggestion:** improve the 3.3 -> 3.4 migration guide's section "[Update package.json commands](https://deploy-preview-877--heuristic-almeida-1a1f35.netlify.app/docs/3.x/upgrade/migration-guides/3.3-3.4#update-packagejson-commands)" to make it more obvious (it looks like it was overlooked in your context).
- update title
- add bullet points: add "type" field, update commands

docs/migration/manual-migration.mdx
real34: **nitpick:** I don't think it's relevant to document a lot of things here. An example is ok, but not more.

**suggestion:** for details I think it's better to link to one (or several) of articles published on the web … this change has been around for 10 years. There are lots of resources. Example: https://blog.logrocket.com/alternatives-dirname-node-js-es-modules/

real34 (changes requested):

docs/migration/manual-migration.mdx
Kagaminara: Added Paul's example and updated the section with links.

docs/migration/manual-migration.mdx
Kagaminara: Applied your comments and moved the section to 3.3->3.4 migration guide.

real34 (approved):
Thanks! :+1:

PR #881, opened 2024-04-22
chore: document rate limiter service

paul-vd:
## What?

This documents the `RateLimiter` and different implementation details.


mGGk-fr:
It should be mentioned in Migration guide on how to switch from legacy method to service usage

docs/02-guides/prevent-excessive-usage-with-rate-limits.mdx
```diff
@@ -0,0 +1,186 @@
...
+import { defineConfig } from "@front-commerce/core/config";
+
+export default defineConfig({
+  // ...other configuration options
+  rateLimiter: {
+    redis: {
+      host: process.env.FRONT_COMMERCE_CLOUD_REDIS_SESSIONS_HOST,
+      port: process.env.FRONT_COMMERCE_CLOUD_REDIS_SESSIONS_PORT || 6379,
+      db: process.env.FRONT_COMMERCE_CLOUD_REDIS_SESSIONS_DB || 2,
+    },
```
real34, line 47: **nitpick:** we can also suggest to extract this if they feel it more convenient

**suggestion:** 
```suggestion
    // feel free to extract this configuration in a separate file to ease maintenance
    redis: {
      host: process.env.FRONT_COMMERCE_CLOUD_REDIS_SESSIONS_HOST,
      port: process.env.FRONT_COMMERCE_CLOUD_REDIS_SESSIONS_PORT || 6379,
      db: process.env.FRONT_COMMERCE_CLOUD_REDIS_SESSIONS_DB || 2,
    },
```

docs/02-guides/prevent-excessive-usage-with-rate-limits.mdx
```diff
@@ -0,0 +1,186 @@
...
+  },
+});
+```
+
```
real34, line 51: **issue:** I think it's important to highlight that we recommend to shate the same Redis connection than sessions, hence our suggestion, but that it's not required.

**suggestion:**  something like that (to be refined as you prefer)

```suggestion
:::tip

We've suggested a configuration which shares the same Redis DB than sessions. While it's not required, we think it makes sense because both share the same kind of responsibilities.

:::
```

docs/02-guides/prevent-excessive-usage-with-rate-limits.mdx
```diff
@@ -0,0 +1,186 @@
...
+For GraphQL resolvers, the rate limiter can be directly applied to individual
+queries or mutations to manage resource consumption effectively.
+
+```ts
+import { createGraphQLRuntime } from "@front-commerce/core/graphql";
+
+export default createGraphQLRuntime({
+  resolvers: {
+    Query: {
+      helloWorld: async (parent, args, context, info) => {
```
real34, line 128: **nitpick:** most of the documentation use a direct context spread, and doesn't document `info` which IMO is too technical for userland apps.

**suggestion:** adapt the signature (and usages below)

```suggestion
      helloWorld: async (parent, args, { user, services }) => {
```

real34 (approved):
It looks good. Only a few nitpicks & minor issues left.

paul-vd:
> It should be mentioned in Migration guide on how to switch from legacy method to service usage

No need, as it's covered by a codemod

PR #887, opened 2024-04-24
docs(remix): adding flash messages related method

mGGk-fr:
# What

This add methods related to flash message in documentation


real34 (approved):
**issue:** just in case, there was this acceptance criteria
> Change for implementers of the Session adapter interface

**suggestion:** add a section in the migration guide mentioning these 2 new methods if one implements the SessionAdapter somewhere. No need of more details and blabla.


Kagaminara (approved):

PR #892, opened 2024-04-26
chore: update all GraphqlModule and GraphqlRuntime docs

paul-vd:
## What?
This documents all the updates for `GraphqlModule` and `GraphqlRuntime` changes

### DIFF
--- a/docs/02-guides/extend-the-graphql-schema.mdx
+++ b/docs/02-guides/extend-the-graphql-schema.mdx
@@ -33,9 +33,7 @@ import { defineExtension } from "@front-commerce/core";
 export default function clickCounter() {
   return defineExtension({
     name: "click-counter",
-    configuration: {
-      providers: [],
-    },
+    meta: import.meta,
   });
 }
 ```
@@ -55,9 +53,11 @@ For that, you can create the GraphQL module definition in the file
 `example-extensions/click-counter/graphql/index.ts` with the following content:
 
 ```typescript title="example-extensions/click-counter/graphql/index.ts"
-export default {
+import { createGraphQLModule } from "@front-commerce/core/graphql";
+
+export default createGraphQLModule({
   namespace: "ClickCounter",
-};
+});
 ```
 
 This is a minimal GraphQL module definition which namespace is `ClickCounter`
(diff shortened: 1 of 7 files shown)


docs/02-guides/extend-the-graphql-schema.mdx
```diff
@@ -98,16 +99,25 @@ We would like to:
 1. add a `clickCounter` field to the existing `Product` type
 2. add an `incrementProductCounter` mutation
 
-For that, you can create the file `extension/click-counter/graphql/schema.gql`:
+For that, you can add the schema the GraphQL module definition file.
```
real34, line 102:
```suggestion
For that, you can add the type definitions as a `typeDefs` value in your GraphQL module definition.
```

docs/02-guides/extend-the-graphql-schema.mdx
```diff
@@ -201,43 +182,47 @@ To learn more about resolvers and their internals, we recommend reading
 
 :::
 
-First, let's update the module definition as follow:
+First, let's update the module definition to add the runtime api:
```
real34, line 185:
```suggestion
First, let's update the module definition to register the runtime API:
```

docs/02-guides/extend-the-graphql-schema.mdx
```diff
@@ -54,10 +52,12 @@ to the extension containing a schema and some resolvers.
 For that, you can create the GraphQL module definition in the file
 `example-extensions/click-counter/graphql/index.ts` with the following content:
 
-```typescript title="example-extensions/click-counter/graphql/index.ts"
-export default {
+```typescript title="example-extensions/click-counter/graphql.ts"
```
real34, line 55: **issue:** I think we should document as we use it in the core. All modules are in an `index.ts` so that the module definition can be in the same dir than runtime and loaders

**suggestion:** restore `graphql/index.ts` that will be more consistent with the `runtime.ts` introduced below.

_same for other occurrences_

docs/03-extensions/contentful/how-to/add-x-ray-compatibility.mdx
```diff
@@ -176,11 +176,56 @@ You need to register your defined `StorefrontContentType` via the Contentful
...
+  `,
+});
+```
+
+Then you can add your `GraphqlRuntime` definition to register your
+`StorefrontContentType`:
+
+```js
+import { createGraphQLRuntime } from "@front-commerce/core/graphql";
+
```
real34, line 227: **nitpick:** the `resolvers` below is undefined

**suggestion:** remove `resolvers`, or add an `import resolvers from "./resolvers"` here

real34 (approved):
Thanks for this update.

PR #889, opened 2024-04-25
chore: document refactored sitemap api

paul-vd:
## What?

This documents the refactoring of the sitemap service api

## Preview

- https://deploy-preview-889--heuristic-almeida-1a1f35.netlify.app/docs/3.x/guides/customize-the-sitemap

## Unrelated

I made the requirements for extending DI type declaration more clear by expanding the install command for `@front-commerce/types` in a codeblock ([internal reference](https://frontcommerce-hq.slack.com/archives/C03C2T9ADLY/p1713536856932899?thread_ts=1713535969.091409&cid=C03C2T9ADLY))

PR #893, opened 2024-04-26
3.4: release announcement

real34:
This MR contains the release announcement for the 3.4 release.

**Status:**
- the "Other changes > Bug Fixes" section needs some more cleanup
- I must create an image (or table) with numbers of time improvements for Vite actions

**[Preview](https://deploy-preview-893--heuristic-almeida-1a1f35.netlify.app/changelog/front-commerce-3.4-2.29)**


changelog/front-commerce-3.4-2.29.mdx
```diff
@@ -0,0 +1,471 @@
...
+changes to their entrypoints in order to benefit from the low-level improvements
+brought by Front-Commerce.
+
+In Front-Commerce 3.4, we introduce a simpler API. We have extracted our default
+logic into the `@front-commerce/remix` package, **allowing projects to use the
+default configuration with just one line of code.**<br/> This eliminates the
+need for manual entrypoint modifications and simplifies the update process.
+
+Of course, projects still have the flexibility to "eject" from the default
+configuration for more advanced use cases.
```
paul-vd, line 262: I think this is only partially true, although it would be ideal to have this also for `entry.server.tsx` and `entry.client.tsx` it's currently only true for `entry.worker.ts`, it might be good to mention that here, and to note that we aim to do the same for server and client entry points.

I would even argue that we should see if they do not exist, that we automatically inject them, so to eject from them, they need to be created in the userland code. But no need to mention that here.

changelog/front-commerce-3.4-2.29.mdx
```diff
@@ -0,0 +1,471 @@
...
+and `@front-commerce/magento2` have has been updated to use this new API.
+
+:::note Previous work
+
+This API was inspired by the work of
+[@balavishnuvj/remix-seo](https://www.npmjs.com/package/@balavishnuvj/remix-seo)
+and the [NASA GCN fork](https://github.com/nasa-gcn/gcn.nasa.gov). Thanks to
+them for their contributions!
+
+:::
```
paul-vd, line 204: @real34 if we get https://github.com/front-commerce/developers.front-commerce.com/pull/889 reviewed and merged, then you could possibly use the artwork there to add a brief explanation of the Sitemap Service api?

https://deploy-preview-889--heuristic-almeida-1a1f35.netlify.app/docs/3.x/guides/customize-the-sitemap#creating-and-registering-a-custom-fetchers

changelog/front-commerce-3.4-2.29.mdx
real34: I reworded as follow:
> In Front-Commerce 3.4, we introduce a simpler API for some of them. We have
extracted our default logic for the express server and the service worker into
the `@front-commerce/remix` package, **allowing projects to use the default
configuration with just one line of code.**<br/> This eliminates the need for
manual entrypoint modifications and simplifies the update process.

changelog/front-commerce-3.4-2.29.mdx
real34: I duplicated it for the changelog entry. This way, the doc could evolve in the future and both are independent.

PR #920, opened 2024-07-02
docs(adyen): adding note about store configuration needed for adyen

mGGk-fr:
# What

This adds a a note about store configuration needed for Adyen payment to work properly

### DIFF
--- a/docs/03-extensions/adyen/how-to/front-commerce.mdx
+++ b/docs/03-extensions/adyen/how-to/front-commerce.mdx
@@ -10,6 +10,14 @@ sidebar_position: 3
 
 ## Front-Commerce configuration
 
+:::tip
+
+Please be sure to configure your store's URL properly in the `app/config/stores.js` file.
+
+Misconfiguration of the store's URL can lead to issues with the Adyen payment methods.
+
+:::
+
 [After installing the Adyen package](/docs/3.x/extensions/adyen/#installation),
 you need to enable the corresponding extension. For that, you need to tweak your
 `front-commerce.config.ts` file like:


real34 (approved):

docs/03-extensions/adyen/how-to/front-commerce.mdx
```diff
@@ -10,6 +10,14 @@ sidebar_position: 3
 
 ## Front-Commerce configuration
 
+:::tip
+
+Please be sure to configure your store's URL properly in the `app/config/stores.js` file.
+
+Misconfiguration of the store's URL can lead to issues with the Adyen payment methods.
+
+:::
```
paul-vd, line 19: It would be good to link to the docs or show an example of how it should be configured, as this is all good and well, but is not clear what `configure your store's URL` is?

paul-vd (changes requested):

docs/03-extensions/adyen/how-to/front-commerce.mdx
mGGk-fr: Except the page on the core API didn't found something to explain to the customer the full procedure to setup stores, that's why I haven't linked this to anything

real34: **nitpick:** It's also not very clear what issues can happen.

**suggestion:** 
```suggestion
:::tip

Please be sure to configure your store's URL properly in the `app/config/stores.js` file.

Misconfiguration of the store's URL can lead to incorrect redirections for Adyen payment methods requiring 3DS authentication.

:::
```

real34 (approved):

docs/03-extensions/adyen/how-to/front-commerce.mdx
mGGk-fr: It doesn't causes incorrect redirections, it causes fail to process requests since shop origin for 3DS request doesn't match origin for payment validation call. I think we shouldn't mention 3DS since a valid configuration should be set-up to ensure eveything works properly.

paul-vd: > Except the [page on the core API ](https://developers.front-commerce.com/docs/3.x/api-reference/front-commerce-core/config#stores)didn't found something to explain to the customer the full procedure to setup stores, that's why I haven't linked this to anything

That is the problem, and hence why this issue can occur, insufficient documentation. hence my comment.

mGGk-fr: I think it deserve a whole section about different configurations / configuration files in the `app/config` folder but it should be part of another issue

PR #967, opened 2024-10-21
Upgrade Docusaurus to v3 and Update Dependencies

paul-vd:
The primary change in this update is the upgrade of Docusaurus from version 2 to version 3, along with updates to other dependencies to their latest versions. Additionally, the Node.js version requirement has been updated to 20.12, and the package manager version for pnpm has been updated to 9.12.

The codebase has been refactored to remove unnecessary React imports, as React 17+ does not require explicit import of React in every file. Type annotations have been adjusted to use `React.ReactNode` instead of `ReactNode` directly.

Documentation links have been corrected and updated to reflect the new structure and paths introduced with the Docusaurus v3 upgrade. These links were hash links were broken since migration to docusaurus but they went undetected because the docosaurus v2 did not alert on them. There were also new v3 docs which had some broken links since refactoring of docs, and same reason, they were undetected in docosaurus v2.

Overall, these changes aim to modernize the project setup, improve compatibility with the latest tools, and ensure the documentation remains accurate and functional.


real34 (approved):
Thanks for the cleanup.

It looks good in preview and I haven't detected anything shocking from a quick review, so I'm OK to ship-it.

It must however be done on `main` right away (and merged back into `next`) IMO to prevent issues with hotfixes landing in `main`.

